PR #23, opened 2018-08-14
fix(upload)Input expects string for size attribute

jsarman:
Warning: Failed prop type: Invalid prop `size` of type `number` supplied to `Input`, expected `string`.
    in Input (created by FilePicker)
    in FilePicker (created by FilePickerBtn)


jsarman:
The problem is that React expects Strings on all input attributes, but in this case it has to be a number. `<input size="number"> `  I think it is better to treat it as a number as it propagates down to the actual input and then cast it to a String to please React.

jsarman:
I agree with that

PR #27, opened 2018-10-09
Pagination

KaseyPowers:
Starting PR to discuss changes/improvements as pagination is migrated to availity-react


KaseyPowers:
changes so far from the internal pagination: 
- made the view pages logic a little easier to follow (hopefully)
- added first/last buttons
- changed the button props syntax so that true = default text and can still accept custom strings

Ideas for basic pagination:
- drop the total items and itemsPerPage and just use a page and totalPages params
- some more configurations for when buttons are used such as:
  - only showing first/last if the total pages to display is larger than what fits in the padding
  - not showing buttons instead of disabling them?

KaseyPowers:
Can now accept the pageCount as well as using totalCount/itemsPerPage logic (pageCount will be used if both defined)

robmcguinness:
unit tests?

KaseyPowers:
Currently WIP, started the PR early to get some feedback on the changes from the original while working on it.

KaseyPowers:
Reworked default Pagination a bit, removing `pageOnlyOptions` and adding support for the items props to be a function. (including async function)
-Some changes possible on top of this with the new PaginationControls is making more controlled/uncontrolled aspects beyond just page. (such as itemsPerPage)
-AsyncPagination can be changed to AvResourcePagination or something similar to similarly wrap our Api calls for Pagination but letting the default handle more of the async aspects now.

KaseyPowers:
- Created AvResourcePagination to wrap Pagination with our api specific Api logic.
- Added ItemsExtraArgs as an object that gets passed as the third argument to any items calls, and to reset the items/page if it changes.

robmcguinness:
@KaseyPowers can you review the preview on Netlify and see if the docs look right to you? I merged from `master` and applied the Storybook 4 upgrade.

jest.setup.js
```diff
@@ -2,3 +2,13 @@ import Enzyme from 'enzyme';
 import Adapter from 'enzyme-adapter-react-16';
```
robmcguinness, line 2: Can this be removed now that we have switched to `react-testing-library`?

GoPro16:
I feel like this library is making the pagination more complicated than it has to be. For just about all use cases, controls/select options will all be under a pagination `HOC` and in this case those components are okay. However, there are much cleaner ways to implement this with hooks and context api. Take the below commit I just made: 

[code](https://github.com/Availity/availity-react/blob/5725296d279d35a6b8fbe7a3e096e8b80094fa61/packages/docs/stories/pagination.stories.js#L60)
```javascript
<Pagination items={items}>
        <PaginationContent component={Component} />
        <PaginationControls directionLinks />
</Pagination>
```

You can pull down the repo and see what I did in detail but this makes it really simple for adding into any project. Yes there are features missing from it like the async pagination with the `items` prop as a function but this is a minor example that hopefully can be worked off of. I mainly created a new repo as well because this is way behind master and was giving my two sense on the current impl.

Using the `useContext` hook, and the context api for that matter, allows developers to easily access all the data they need from the pagination component if they want to make custom controls, etc. without having to access via a render prop.

We can discuss this tomorrow and come to an agreement on what steps can be taken to get this merged in. All the code that was written for the async is all good and the helper components are also a plus. The idea is to keep the `Pagination` component as a dumb component and passing the props from somewhere else in. The only thing it can manage is the `page` since its paginating the data.

jest.setup.js
GoPro16: Yes the whole jest setup can be removed as well as the tests that are testing if the `propTypes.required` works in which it should already be expected.

KaseyPowers:
I agree about the context situation @GoPro16 . I was initially simplifying out the pagination from av-react into individual components giving more control on where the controls are placed. I did realize though that the final pagination component left was for the most part more difficult to use than just doing that logic yourself. 

I had talked to @robmcguinness about moving the logic into a context provider but a direction wasn't decided. 

It seems from your PR #57 that some of the async logic from here and the more flushed out controls could be mixed into what you have to be overall more effective?

GoPro16:
@KaseyPowers yes the async logic can definitely be brought over into this components. I was trying to devy out all the props to the components that should be using them rather than having them all placed on the top level. For now I brought over simple ones but we can always add the rest in when this gets merged. 

The last thing for this to get merged at minimum is the `AsyncPagination` component. I am working on this now and all the logic that you have built is just being moved into the new `Pagination` component. 

After that it should be easy to add all the additional features you wanted to put on top of the component as it all should be cleaned up and easier to see what component can take what feature.

PR #29, opened 2018-10-31
feat(av-loadable): init work on common availity loadable

KaseyPowers:
Wanted to share some initial work on creating a common Loader component for https://github.com/jamiebuilds/react-loadable
using our block-ui

### DIFF
--- a/packages/av-loadable/AvLoadable.js
+++ b/packages/av-loadable/AvLoadable.js
@@ -1,0 +1,16 @@
+import Loadable from 'react-loadable';
+
+import Loading from './Loading';
+
+function AvLoadable(opts) {
+  return Loadable(
+    Object.assign(
+      {
+        loading: Loading,
+      },
+      opts
+    )
+  );
+}
+
+export default AvLoadable;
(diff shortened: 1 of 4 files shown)


TheSharpieOne:
Note: jamiebuilds/react-loadable doesn't currently support IE 11 out of the box since it uses `Object.assign` in it's dist file :(
You would be required to add a polyfill for `Object.assign` or run the package through babel.

https://github.com/jamiebuilds/react-loadable/pull/85
https://github.com/jamiebuilds/react-loadable/pull/144#issuecomment-422606007

KaseyPowers:
I was unaware of this, I must be pollyfilling that in the project we are using this in. I know I added more than regular for other issues but forget if this was one of them. 

There is the conditional polyfill that could be done manually and added to this? 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign#Polyfill

TheSharpieOne:
Closed in favor of `React.lazy`?

robmcguinness:
@TheSharpieOne yes.

PR #36, opened 2019-01-05
fix(reactstrap-validation-select): allow component proptype for label. Fixes #35

### LINKED ISSUE #35: AvResourceSelect Label prop is string only
When using AvResourceSelect the label prop is required to be a string, should be a string or node.

context: want to add a required * with a span around it for styling.

GoPro16:
Fixes Issue #35.


packages/reactstrap-validation-select/AvResourceSelect.js
```diff
@@ -15,7 +15,7 @@ class AvResourceSelect extends Component {
     }).isRequired,
     getResult: PropTypes.oneOfType([PropTypes.string, PropTypes.func]),
     delay: PropTypes.number,
-    label: PropTypes.string,
+    label: PropTypes.oneOfType([PropTypes.string, PropTypes.node]),
```
TheSharpieOne: `PropTypes.node` includes `PropTypes.string`, so just setting it to `PropTypes.node` would be good enough.

PR #37, opened 2019-01-05
Proposal to Update README Props Formatting.

GoPro16:
Currently the props formatting is inline sentences. Simply adding a table format makes this more readable. Added Fullscreen link to the top of the README in case users are viewing storybook and want to go to the Github link where they can see the fullscreen view. 

Additional horizontal rules added to separate the component docs and bumped the font size on the component names.


robmcguinness:
closing. we will refactor docs in another PR.

PR #41, opened 2019-02-01
Feature migration - Adds Components From Av-React. Replaces Enzyme with React-Testing-Library

GoPro16:
Jest has been upgraded to version 24.0.0.

I added to the jest.config.js with the below:
`
setupFilesAfterEnv: [
    'react-testing-library/cleanup-after-each',
  ]
`
[Code Link](https://github.com/Availity/availity-react/blob/feature-migration/jest.config.js#L5)

However, this script is not getting auto applied on cleanup as if you were to remove the below line from `Upload.test.js` it will fail to complete testing because the cleanup script doesn't get applied:
`import 'react-testing-library/cleanup-after-each';`
[Code Link](https://github.com/Availity/availity-react/blob/feature-migration/packages/upload/tests/Upload.test.js#L4)


robmcguinness:
LGTM.

PR #45, opened 2019-02-05
refactor(feature): updated aria attributes

GoPro16:
Updated Aria Attributes for:
- Breadcrumbs: added aria-labels for `a` tags.

- Page Header: Only 1 `h1` tag is allowed per page. We should not be putting the h1 element inside the page header in the case its not the main element of focus.

### DIFF
--- a/packages/page-header/PageHeader.js
+++ b/packages/page-header/PageHeader.js
@@ -30,7 +30,7 @@ const PageHeader = ({
       <div className="d-flex align-items-start">
         <Breadcrumbs crumbs={crumbs} active={appName || children} /> {component}
       </div>
-      <h1 className="h4 page-header page-header-brand" {...props}>
+      <h4 className="page-header page-header-brand" {...props}>
         <div className="page-header-title">
           {!payerId && appAbbr && (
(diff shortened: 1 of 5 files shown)


packages/page-header/PageHeader.js
```diff
@@ -30,7 +30,7 @@ const PageHeader = ({
       <div className="d-flex align-items-start">
         <Breadcrumbs crumbs={crumbs} active={appName || children} /> {component}
       </div>
-      <h1 className="h4 page-header page-header-brand" {...props}>
+      <h4 className="page-header page-header-brand" {...props}>
```
TheSharpieOne: This is just as much as an opinion as the `h1` as setting `h4` assumes that the page already defines `h1`, `h2`, and `h3` before-hand since you shouldn't skip a heading level.

Maybe this tag should be configuration (similar to the `tag` prop in reactstrap). Since this is supposed to be the _page_ header (not a section header), and typically (branding) it is the name of the page header and the most top level header, it could default to `h1` but be overridable for less standard uses elsewhere within the page.

PR #47, opened 2019-02-07
Feature/typescript

GoPro16:
Added Typings for each of the components. This is not a breaking change since if the users are using standard React the index.ts file will never be called. 

I also updated the package.json's of the new components to point to the npm public registry and not artifactory. This probably should have been underneath a different commit but since the changes here are non breaking I added the changes with this branch.

If you have any recommendations please let me know. The idea is to add the typings in here for tsx support in projects.


tsconfig.json
```diff
@@ -0,0 +1,33 @@
...
+    "jsx": "react",
+    "moduleResolution": "node",
+    "forceConsistentCasingInFileNames": true,
+    "strict": true,
+    "noEmit": true,
+    "experimentalDecorators": true,
+    "baseUrl": "./packages",
+    "allowSyntheticDefaultImports": true,
+    "paths": {
+      "@mavaility/app-icon": ["./app-icon/src"],
```
TheSharpieOne: typo

GoPro16: Good catch on that 👍

PR #48, opened 2019-02-07
Breadcrumbs children

jeffski13:
-storybook examples
-unit tests for breadcrumb render prop "renderCustomCrumbContent"
-breadcrumb prop renderCustomCrumbContent docs in readme


packages/breadcrumbs/package.json
```diff
@@ -19,12 +19,11 @@
   "publishConfig": {
     "registry": "https://artifactory.availity.com/artifactory/api/npm/local-npm/"
   },
-  "dependencies": {
-    "prop-types": "^15.5.8"
-  },
   "devDependencies": {
+    "prop-types": "^15.5.8",
```
GoPro16: For Travis to build the project you will also have to commit the package-lock file or else it will throw an error.

jeffski13: right on! I like those changes you suggested! thanks!

TheSharpieOne: This should stay as a dependency and not a devDependency per: https://github.com/facebook/prop-types#how-to-depend-on-this-package

Also it is important to understand that `prop-types` is a legit dependency, it _is_ required to run this package as the code imports prop-types and utilizes it. With this change, if a project failed to include their own prop-types dependency it would break.

jeffski13: Thank you for clarifying! It always seemed like one of those development only things to me, but that makes sense.
Also whats up Evan

jeffski13: hey guys, it looks like every package-lock.json in the project was updated. Do i need to include all of those? or just the package-lock.json's next to the package.json's i modified?

jeffski13:
> LGTM other than @TheSharpieOne 's comment and you still need to update the component for storybook as well.

I made those changes. Im not terribly familiar with travis ci and these build tools. Did I mess something up?

GoPro16:
It looks like the package-lock files are still not correct. What I would do is run `npm run nuke` and then `npm install`. This will remove all the package-locks and node_modules and you can reinstall to get this corrected.

PR #51, opened 2019-02-13
fix: removed package-locks for now

GoPro16:
This build should succeed with the new travis settings.


robmcguinness:
I think this is no longer needed.

PR #52, opened 2019-02-13
fix(reactstrap-validation-date): fix for toLowerCase() error in AvDate.js

joseflores-io:
clicking on the Calendar component after entering a space in AvInput component causes issue

	* added date validation to fix
	* will only setState if input is a valid date


0xjcf:
using the validation from availity-reactstrap-validation

PR #54, opened 2019-02-20
feat: integrated dynamic breadcrumbs into PageHeader

jeffski13:
I modified the Pageheader to use the new dynamic Breadcrumbs.
Props validation added for crumbs and custom children BreadcrumbItems

### DIFF
--- a/packages/breadcrumbs/Breadcrumbs.js
+++ b/packages/breadcrumbs/Breadcrumbs.js
@@ -49,16 +49,40 @@ Breadcrumbs.propTypes = {
   active: PropTypes.string.isRequired,
   emptyState: PropTypes.string,
   children: (props, propName, componentName) => {
-    const prop = props[propName];
-
+    if (!props.children) {
+      return null;
+    }
     let error = null;
-    React.Children.forEach(prop, child => {
-      if (child.type !== BreadcrumbItem) {
+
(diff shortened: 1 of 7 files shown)


packages/breadcrumbs/Breadcrumbs.js
```diff
@@ -49,16 +49,40 @@ Breadcrumbs.propTypes = {
...
-
+    if (!props.children) {
+      return null;
+    }
     let error = null;
-    React.Children.forEach(prop, child => {
-      if (child.type !== BreadcrumbItem) {
+
+    // multi children
+    if (Array.isArray(props.children)) {
```
robmcguinness, line 58: Can we add a corresponding test case using `react-testing-library`?

jeffski13: DONE!

jeffski13:
All i did was update some new unit tests, and now Travis hates me ):

GoPro16:
@robmcguinness  Looks like the build is starting to take longer now since we are adding more components to the story book.  Can you try re-starting the netlify build? I am pretty sure this is the case that is happening right now.

GoPro16:
@jeffski13  review the changes I just made and let me know if you see any issues with it.

PR #56, opened 2019-02-22
Fix/payer logo

### LINKED ISSUE #55: payer-logo images will not always work
payer-logo assumes images will be in one of two places:

https://github.com/Availity/availity-react/blob/master/packages/payer-logo/PayerLogo.js#L7-L8

that is not always the case. ie could be in vault

Proposal: make ajax call to get logo from space configuration and then render it

nylon22:
closes #55 

makes ajax call to slotmachine to get appropriate logo by either the `spaceId` or `payerId` passed in props


TheSharpieOne:
Also, you probably want to track if the payer id or space id change in case a component is reused and not completely unmounted and remounted.
Would you want a way to cache the URLs more globally so that it doesn't trigger more than 1 network requests per payer id / space id in total?
Potentially, showing a list payers and payer logos would fire a new request for every item and when the pagination changes, it will do it again.

I recall now that there was a plan to make vanity URLs for accessing the images, something where `/some/path/payerid/logo.png` would (on the backend) make the call to determine the location of the image and then send the image back (the actual image, not a reference/URL). The underlying image could change, but the vanity URL would remain the same predicable URL and could be used directly in the `img` `src`.

GoPro16:
With React `16.8` being released its a lot easier now to migrate Class Components into Functional ones with `Hooks`. You can easily take what you just added and put it as a Functional Component.

`useState` [Reference](https://reactjs.org/docs/hooks-state.html)
You have 1 piece of state that you need to keep and that is the `URL`. In this case you can create a single piece of state like the below:
```javascript
const [url, setUrl] = useState(null);
```

`useEffect` [Reference](https://reactjs.org/docs/hooks-reference.html#useeffect)
From what your componentDidMount method is doing and from @TheSharpieOne 's comments above you can just use the `useEffectAsync` hook from `@availity/hooks` in order to get your url. You can then add the 2nd argument to the `useEffectAsync` as the `[spaceId, payerId]` and only when those props change will it re-render.
```javascript
useEffectAsync(async () => {
... await stuff
setUrl(url);
},[spaceId,payerId]);
```

nylon22:
The vanity urls have been talked about, but i don't think there's an action plan for doing it. I have a more immediate need. Once that work is done, we can update this component again.

On the subject of caching urls and tracking changes to the payerId/spaceId props - great feedback. I'll update this PR 👍

nylon22:
Any thoughts on how to cache the urls?

Also, I believe the travis build is failing because [these deps](https://github.com/Availity/availity-react/blob/master/package.json#L71-L72) need to be upgraded to `^16.8.3`

GoPro16:
For caching using the `AvSlotmachine` which hits graphQL there is not easy way to know if the data has changed or not. For now we should keep it as is and in the future look at a way we all agree on for data caching. `react-cache` is going to be released some time this year and we did support a`local-cache` in the `sdk-js` but I don't think its good to do that here since we still don't know if there has been an update or not.

I believe you are correct for that. You can change the `react` version in the base `package.json` and it should be okay. We don't publish the main package out so all the config at the base level is for testing deps and all.

packages/payer-logo/PayerLogo.js
```diff
@@ -27,18 +28,66 @@ query($payerIDs: [String!], $types: [String!]){
 }
 `;
 
-export default class PayerLogo extends Component {
-  state = {};
+const getLogoBySpaceId = async spaceId =>
+  new Promise(async (resolve, reject) => {
```
GoPro16: I think we can do away with creating a promise here and just use `async`/`await` syntax. Will reduce all the extra setup for the `resolve/reject`.

packages/payer-logo/PayerLogo.js
```diff
@@ -27,18 +28,66 @@ query($payerIDs: [String!], $types: [String!]){
...
+    }
+  });
 
+const getLogo = (spaceId, payerId) =>
+  new Promise(async (resolve, reject) => {
     try {
       let url;
       if (spaceId) {
-        url = await this.getLogoBySpaceId();
+        url = await getLogoBySpaceId(spaceId);
```
GoPro16: There should be a way you can just create 1 function out of this and pass in a simple object with params so we don't have a lot of duplicate logic in both `GET` methods.

TheSharpieOne:
https://github.com/Availity/sdk-js/tree/master/packages/keyv-local-sync
Supports promises so that if you end up getting a list which all of the same payers they are get the same promise and when that promise resolves they all get the response. Many components/renders, 1 request.
It also supports TTL to ensure the cache doesn't live too long.
react-cache doesn't solve the issue of if the cache is invalid such as someone changed the logo on the server.
Plus, if the logo changes while a user is in the middle of a session, I wouldn't really expect them to see the new logo along with the old one (given that the old one may still be on the page) I would expect them to have to reload to get the new one (until server push is allowed).

packages/payer-logo/PayerLogo.js
nylon22: please see 337253dfc06c9b6fbdedcdbd31b782e696f64f55

packages/payer-logo/PayerLogo.js
nylon22: please see 337253dfc06c9b6fbdedcdbd31b782e696f64f55

GoPro16:
LGTM, looking into why the `useEffectAsync` during testing is causing that invariant issue.

PR #57, opened 2019-02-22
Pagination/hooks

### LINKED ISSUE #64: remove react-dom/test-utils from PayerLogo
Test utitlies should not leak in components.

https://github.com/Availity/availity-react/blob/master/packages/payer-logo/PayerLogo.js#L2
```js
import { act } from 'react-dom/test-utils';
```

GoPro16:
This PR contains 3 components & 1 Hook:
- Pagination ( Provider)
- PaginationControls
- PaginationContent
- usePagination

Todos -
- [x] Unit Test
- [x] Async Pagination
- [x] Better StoryBook examples
- [x] Typings

closes #27 
closes #64


packages/pagination/PaginationContent.js
```diff
@@ -0,0 +1,38 @@
...
+        page.items.map((value, key) => {
+          if (!value.key) {
+            // eslint-disable-next-line no-console
+            console.warn(
+              "Warning a Pagination Item doesn't have a key:",
+              value
+            );
+          }
+
+          return <Component key={value.key || key} {...value} />;
```
TheSharpieOne: Maybe `...rest` the props and also spread them on here. It would help avoid a component in the middle just to set additional props?

packages/pagination/PaginationControls.js
```diff
@@ -0,0 +1,58 @@
...
+              active={current === pageNumber}
+              onClick={() => setPage(pageNumber)}
+            >
+              <PaginationLink>{pageNumber}</PaginationLink>
+            </PaginationItem>
+          ))}
+          {directionLinks ? (
+            <PaginationItem
+              disabled={current === pages.length}
+              onClick={() =>
```
TheSharpieOne: `PaginationItem` output's an `li` which is a non-interactive element, you probably want to change the `tag` or move the `onClick` to the `PaginationLink` which will output a `button` (and, if you do that, also add `type="button"` to it so that it doesn't trigger submits if it appears within a form).

packages/pagination/Pagination.js
```diff
@@ -0,0 +1,108 @@
+import React, { useState, useContext } from 'react';
+import PropTypes from 'prop-types';
+import isFunction from 'lodash/isFunction';
+import sortBy from 'lodash.sortby';
```
TheSharpieOne: you are using components form the main lodash lib above (`lodash/isFunction`), but here you install and use the specific lib (`lodash.sortby`). Should be consistent.

GoPro16:
> Local sorting and searching seems like a red herring, it requires all of the data to be loaded into memory to work as an end-user would expect and I wouldn't expect developers to load all of the data in memory.
> 
> I would imagine that most of the use-cases at Availity involve remote data. While it is nice to have the lower level functionality it would be better to provide that Async component with the initial offering or it might become the wild west of paging remote data (there is a page currently in the portal which loads all of the data from all of the pages and since the API limits the `limit` to 50, it happens 50 at a time... for 10k+ records... in serial... blocking the UI until everything is loaded. IDK why they did that, but I would imagine it had something to do with being able to search the data in memory).
> 
> If you haven't already, take a look at the pagination in av-react. The async there is opinionated to work with Availity's SDK-js and for an Availity library, I think that's a good opinion. KC tried to make several enhancements which were nice, but I think at the bare minimum, something to handle remote data is needed.
> 
> Also, maybe Pagination can also accept a FaaC where the function is just called with the context value?

Love the

GoPro16:
Miss clicked the reply button 🤦‍♂️ 

Love the feedback. Yeah I had previously built a `react-bootstrap-pagination` in a private repo of mine that the search and filtering was good for because all the data was already sent to that component for the use case it was serving. In here you are correct its not the case and to make things simpler its better to remove those props and just act as a pager for the data. Adding an Async Pagination Wrapper would not be hard as its just a `HOC` ontop of `Pagination` rather than creating as a seperate provider.

Going to make all those changes you suggested as well as I don't disagree with any of those.


 
> Also, maybe Pagination can also accept a FaaC where the function is just called with the context value?

Not sure if you are reffering to the `usePagination` hook as it already returns all the context values in there. I need to update the README but I also am exported the `PaginationContext` from the component so a Class component can still use the `Component.contextType = PaginationContext`.  I know you said maybe but want to confirm what you a referring to there and probably can add that in as well.

TheSharpieOne:
For the FaaC, I mean that the component which render's Pagination cannot access the context since the context is only available to `Pagination`'s children. Without FaaC, there would be no way for the component which render's Pagination to get access to the data within the context. You force legit components to be children in order to access the data. This is fine and probably enforces component code but seems like overkill for some extremely simple views.

Example:

You can certainly have to create a new component which would have access to the data via `usePagination`

```jsx
const MyItems = () => {
  const { page } = usePagination();
  return page.items.map((item, i) => <div>{page.lower + i}: {item.value}</div>
}

const MyList = () => {
  // cannot do `usePagination` here as this component doesn't have access to the pagination context, only children of Pagination have access.
  return <Pagination>
    <MyItems />
  </Pagination>;
}
```

But it would be nice to also just pass in a function which can get access to the same data that `usePagination` would return without needing to make another component.

```jsx
const MyList = () => {
  // cannot do `usePagination` here as this component doesn't have access to the pagination context, only children of Pagination have access.
  return <Pagination>
    {page => page.items.map((item, i) => <div>{page.lower + i}: {item.value}</div>}
  </Pagination>;
}
```
---
EDIT:
I guess with `PaginationContext` you could do something like:

```jsx
const MyList = () => {
  // cannot do `usePagination` here as this component doesn't have access to the pagination context, only children of Pagination have access.
  return <Pagination>
    <PaginationContext.Consumer>
      {page => page.items.map((item, i) => <div>{page.lower + i}: {item.value}
</div>}
    </PaginationContext.Consumer>
  </Pagination>;
}
```
Right?

GoPro16:
@TheSharpieOne  okay that makes sense. And yes both of your examples are valid. The idea for the `Consumer` is just so that it can be called any where nested in the `Pagination` Provider and not just the first child of it.

GoPro16:
@KaseyPowers I kept all of the props on here as simple as possible for now and then once merged you can go in and add more props in as you see fit.

packages/pagination/src/Pagination.js
```diff
@@ -0,0 +1,97 @@
...
+
+/**
+ * Pagination Provider that takes in an async function or items and provides a list of items
+ * @param {Function|Object} items - The Items or function to use for pagination
+ * @param {number} itemsPerPage - The items to show per page
+ */
+const Pagination = ({ items: theItems, itemsPerPage, children }) => {
+  const [currentPage, setPage] = useState(1);
+  const [pageData, setPageData] = useState({
+    total: theItems != null ? theItems.totalCount : 0,
```
TheSharpieOne: when `theItems` is a function, it will not be null and this will return `theItems.totalCount` which would be `undefined`. Not sure if that is intentional.

packages/pagination/src/Pagination.js
```diff
@@ -0,0 +1,97 @@
...
+    // Get index of item at the start of the currentPage
+    const lower = currentPage === 1 ? 1 : (currentPage - 1) * itemsPerPage + 1;
+
+    // Get the index of the item at the cut-off of the itemPerPage Count
+    const upper =
+      items.length - currentPage * itemsPerPage > 0
+        ? itemsPerPage * currentPage
+        : items.length;
+
+    // Slice that data if it was given from a function since we don't know if its returning all or not
```
TheSharpieOne: This comment makes it sounds like you want to slice the data if it was returned from a function `theItems`, but the it ends up slicing only if `theItems` is not a function. Either the comment is misleading or the ternary is backwards.

packages/pagination/src/Pagination.js
```diff
@@ -0,0 +1,97 @@
...
+    // todo - add prop if needed to handle this
+    const pageData = isFunction(theItems)
+      ? items
+      : items.slice(lower - 1, upper);
+
+    // Get page numbers
+    const pages = Array.from(
```
TheSharpieOne: Going to need a polyfill if you want to use `Array.from`.

packages/pagination/src/Pagination.js
```diff
@@ -0,0 +1,97 @@
...
+
+    setPageData({
+      total: totalCount || items.length,
+      pages,
+      page: { number: currentPage, items: pageData },
+      lower,
+      upper,
+    });
+
+    toggleLoading();
```
TheSharpieOne: Since this function is potentially async, this statement may get triggered when loading has already been toggled off. You may want to support passing a param to the toggle function to explicitly set the value (like some other hook library toggles support) or wrap this in a condition to only run if loading is true.

packages/pagination/src/PaginationContent.js
```diff
@@ -0,0 +1,46 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Util } from 'reactstrap';
+import BlockUI from 'react-block-ui';
```
TheSharpieOne, line 4: may want to include the CSS for BlockUI (though, I'm not 100% since this will be a library and it would require the end-developer's build tool to be able to pick up CSS imports from node_modules)

packages/pagination/src/PaginationContent.js
```diff
@@ -0,0 +1,46 @@
...
+  loadingMessage,
+  itemKey,
+  loader = false,
+  ...rest
+}) => {
+  const { page, loading } = usePagination();
+
+  return (
+    <BlockUI
+      data-testid="pagination-content-con"
```
TheSharpieOne, line 19: is it ok to have `data-testid` here? Seems like testing spilling over.

packages/pagination/src/PaginationContent.js
```diff
@@ -0,0 +1,46 @@
...
+import PropTypes from 'prop-types';
+import { Util } from 'reactstrap';
+import BlockUI from 'react-block-ui';
+import { usePagination } from './Pagination';
+
+const PaginationContent = ({
+  component: Component,
+  loadingMessage,
+  itemKey,
+  loader = false,
```
TheSharpieOne: Why not `defaultProps`?

packages/pagination/src/PaginationControls.js
```diff
@@ -0,0 +1,73 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Pagination, PaginationItem, PaginationLink } from 'reactstrap';
+import { usePagination } from './Pagination';
+
+const PaginationControls = ({
+  directionLinks = false,
```
TheSharpieOne: Why not `defaultProps`?

packages/pagination/src/Pagination.js
TheSharpieOne: Similar to this comment above where multiple requests are in flight (such as the person hitting the next page twice before the first request finishes or just clicking page numbers before each request finishes), how does this handle responses out of order (such as the first request finishing after the second request)?
Usually this causes the wrong page to display. Say the user clicks page 2 and then page 3. The page 3 request finishes and displays the page 3 results, but then the page 2 request finishes... now `currentPage` is still 3, but `pageData` is the items for page 2.
You may want to cancel pending requests before making new ones or just keep track of the most recent request and only care about its response.

packages/pagination/src/Pagination.js
GoPro16: Yeah the idea here was that if the user is providing `theItems` as a function and we are passing the currentPage to it as an arg then the user is in control of the data returned from the function. If the user is passing in an array then its determined to paginate that list and not recalling the function.

packages/pagination/src/PaginationContent.js
GoPro16: I was torn between which one to use, as inlining them with the prop args sometimes is easier to read then going to defaultProps declaration and matching the key value to the prop arg. I will refactor it to just use the defaultProps instead.

packages/pagination/src/PaginationControls.js
GoPro16: Same as comment above.

packages/pagination/src/PaginationContent.js
GoPro16: `data-testid` is just a simple prop for `react-testing-library` to find components in the DOM Tree via `getByTestId` rather than by querySelector or Text.  This way if any of the underlying markup or classes change this will always be the sole identifier of the component for testing.

packages/pagination/src/Pagination.js
GoPro16: No that was unintentional. Adding the `isFunction` check on there as well. 🐛 🔨

packages/pagination/src/Pagination.js
TheSharpieOne: `Slice that data if it was given from a function`

GoPro16: Read over that comment way to quickly, changing that. 👓

packages/pagination/src/Pagination.js
GoPro16: I just added in the latest commit the `localstorage-core` to handle the current page and keep the component in a loading state until the latest page is rendered. We can potentially add the fetch abort options into this. I have seen some people creating a hooks that handle all of that for you and it would be interesting to see the `api-axios` resource in something like the below down the road.

 ```javascript
const [resource,getResource] = useResource(AvOrganizationResource)`
``` 

Inspiration: https://github.com/schettino/react-request-hook

TheSharpieOne: I don't think you need the localstorage stuff. Firstly, it's global (at least all of the pagination components will share the same store and thus the same `current-page`). Not a big deal if there is only ever 1 pagination component rendered at a time.
Second, I believe the values you end up comparing will always match. `currentPage` is scoped out-side of the `useEffectAsync` function and will always be the most recent page. `avLocalStorage.set('current-page', currentPage);` is the first thing to run, so in the most recent/latest time it runs, it will set the value in local storage to the most recent value, which as mentioned before will be the same value used when comparing the older responses in `if (!isEqual(avLocalStorage.get('current-page'), currentPage))`
Since `currentPage` is a primitive, you can create a copy of `currentPage` within the `useEffectAsync` function (something like `const thisPage = currentPage`) and then when `currentPage` changes and is referenced/accessed later it will be the most recent page and `thisPage` will be the page at the time the `useEffectAsync` function ran.

GoPro16: Due to how to the digest cycle works for hooks. Referencing a variable that is called later will actually not take the latest value and instead will just use the current value at the time of the update.

Would it be better then to add a constructor onto the `localStorage` object and optional pass in a prefix or hash for that particular instance. This way the items used in that local storage instance are always guaranteed to be unique?

GoPro16:
> please remove pollyfill

Refactored to use `Array.apply` instead.

packages/pagination/src/Pagination.js
TheSharpieOne: I was mostly talking about variable scoping in javascript, but if hooks do something which make javascript not lexically scoped then that's cool.

PR #58, opened 2019-02-25
Feature/icon

GoPro16:
Adds Icon Component that uses classes from the [availity-uikit](http://availity.github.io/availity-uikit/v3/icons).


GoPro16:
I updated the `Icon` to now just take in the extended class name of the size. Until we update the `availity-uikit` package to better size name conventions I think we should at least support all these until then.

packages/icon/Icon.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const Icon = ({ name, size, ...rest }) => (
+  <i className={`icon icon-${name} ${size ? `icon-${size}` : ''}`} {...rest} />
```
TheSharpieOne: Should we merge in prop className? Are there any a11y attributes which would be good to default (maybe `aria-label={name}`)?

packages/icon/test/Icon.test.js
```diff
@@ -0,0 +1,24 @@
...
+    const { getByTestId } = render(<Icon data-testid="icon" name="home" />);
+
+    const icon = getByTestId('icon');
+
+    expect(icon).toBeDefined();
+  });
+
+  test('should render given a size', () => {
+    const { container } = render(
+      <Icon data-testid="icon" name="home" size="7" />
```
TheSharpieOne: Need to update this for your recent change

packages/icon/README.md
```diff
@@ -0,0 +1,28 @@
...
+```
+
+### Usage
+
+```javascript
+import React from 'react';
+import Icon from '@availity/icon';
+// ...
+<Container>
+    <Icon name="home" size="3" />
```
TheSharpieOne: Need to update this for your recent change

packages/icon/test/Icon.test.js
TheSharpieOne: hmmm... how did this test pass in travis given that recent change which makes the size value more literal. It would have output `icon-7` which is not `icon-5x`

packages/icon/Icon.js
GoPro16: Yeah we can merge those in. And then set the default aria label to `name` of the icon unless specified otherwise.

packages/icon/test/Icon.test.js
GoPro16: Ahh. It looks like it returns null and not undefined. Misguiding

PR #59, opened 2019-02-25
fix(docs): static file path was not being pulled correctly

GoPro16:
The `payer-logo` component pulls static url's for the logos in which they are not present in the build folder due to the exclusion of the `./static` folder.


GoPro16:
@robmcguinness  can you please re-run the netlify build when you get a chance. For some reason the preview isn't building. Otherwise this can be merged and deployed as its not effecting anything.

PR #61, opened 2019-02-26
fix(payer-logo): correct path from payerID query

nylon22:


GoPro16:
Can you write a quick unit test to make sure the logo is getting populated in the `<img /≥`. I ran the current test suite off master and they all succeed.

nylon22:
that is because [the url it defaults to](https://github.com/Availity/availity-react/blob/master/packages/payer-logo/PayerLogo.js#L66) is the same as the [mocked resolved value](https://github.com/Availity/availity-react/blob/master/packages/payer-logo/tests/PayerLogo.test.js#L26) in the test.

I could change the resolved value and write an additional test checking both the default logic and the logic for handling if it gets a url back. that good? @GoPro16

GoPro16:
Hmm, okay we should definitely change that then so we can create a unit test to make sure no future changes break this.

GoPro16:
LGTM.

PR #63, opened 2019-02-26
fix(reactstrap-validation-date): fix distance validation

KaseyPowers:
fix to work when the start name uses lodash syntax (ex. `date.start`).

### DIFF
--- a/packages/reactstrap-validation-date/AvDateRange.js
+++ b/packages/reactstrap-validation-date/AvDateRange.js
@@ -205,16 +205,19 @@ export default class AvDateRange extends Component {
     }
   };
 
-  validateDistance = (end, values) => {
-    if (values[this.props.start.name] && end && this.props.distance) {
-      const start = moment(new Date(values[this.props.start.name]));
+  validateDistance = end => {
+    const start = this.context.FormCtrl.getInput(
+      this.props.start.name
+    ).getViewValue();
+    if (start && end && this.props.distance) {
+      const mStart = moment(new Date(start));
       const mEnd = moment(new Date(end));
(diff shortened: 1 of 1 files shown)


GoPro16:
Can we make sure we have a unit test that passes this case.

packages/reactstrap-validation-date/AvDateRange.js
```diff
@@ -205,16 +205,19 @@ export default class AvDateRange extends Component {
     }
   };
 
-  validateDistance = (end, values) => {
-    if (values[this.props.start.name] && end && this.props.distance) {
-      const start = moment(new Date(values[this.props.start.name]));
+  validateDistance = end => {
```
GoPro16, line 208: Let's unit test this.

KaseyPowers: Since this is currently on snapshot testing, and you mentioned that its due for a refactor/cleanup, I figured unit testing would be a part of that push?

GoPro16: Right, eventually we will have to change this component to a 508 Compliant Date Picker or refactor it so that it will be 508 compliant. At least for all the changes we start making in these repos there should be a unit test for each bug fix because if this issue came up then its apparent that we didn't cover all of the code with the current tests.

We are going to push to get away from snapshot testing. `react-testing-library` helps with rendering and getting items from the virtual DOM as well as triggering different events a lot better than way `react-test-renderer` and `enzyme` can do. We should be testing that certain events are happening and the user can activate certain events rather than testing if a component looks like its snapshot since it can change from the different classes or underlying components.

PR #66, opened 2019-02-28
fix(payer-logo): send clientId in request to slotmachine

nylon22:
BREAKING CHANGE: clientId prop is required. it is needed in order to
call slotmachine

### DIFF
--- a/packages/payer-logo/PayerLogo.d.ts
+++ b/packages/payer-logo/PayerLogo.d.ts
@@ -1,7 +1,7 @@
 export interface PayerLogoProps {
     spaceId?: string;
     payerId?: string;
-    clientId?: string;
+    clientId: string;
 }
 
 declare const PayerLogo: React.FunctionComponent<PayerLogoProps>;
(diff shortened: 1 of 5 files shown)


packages/payer-logo/PayerLogo.d.ts
```diff
@@ -1,8 +1,9 @@
 export interface PayerLogoProps {
     spaceId?: string;
     payerId?: string;
+    clientId?: string;
```
GoPro16: If `clientId` is required then you don't have to add the `?` there. `?` denotes an optional variable.

nylon22: please see ec06bfa499f0f29937a48e5454b389dcc7be8eec

i have not delved into typescript yet 😄

GoPro16: You're good, its new to me as well. Adding typings in is at minimum what we can do right now for `tsx` peeps.

PR #67, opened 2019-03-01
feat(feedback): refactored feedback component with more options

GoPro16:
BREAKING CHANGE: The feedback modal has a new look and could breaking existing tests with snapshots.


packages/breadcrumbs/Breadcrumbs.js
```diff
@@ -21,23 +22,26 @@ class Breadcrumbs extends React.Component {
...
+        <a aria-label="Home" href="/public/apps/dashboard">
+          Home
+        </a>
+      </BreadcrumbItem>
+      {crumbs && crumbs.length > 0 && crumbs.map(renderBreadCrumb)}
+      {children}
+      <BreadcrumbItem active>{active || emptyState}</BreadcrumbItem>
+      {feedback && (
+        <Feedback
+          appName={feedback.appName || active}
```
sanelo: Is 'active' the right backup here? Im assuming 'active' is the name of the active page, not necessarily the 'appName'.

packages/docs/stories/feedback.stories.js
```diff
@@ -20,8 +20,14 @@ storiesOf('Actions|Feedback', module)
     <Feedback
       appName={text('Application Name', 'Payer Space')}
       prompt={text('Prompt')}
-      formOptions={{
+      formProps={{
         additionalComments: boolean('Additional Comments', false),
+        aboutOptions: boolean('About Options', false) && [
+          { label: 'The Payer Space', value: 'space' },
+          { label: 'Applications', value: 'applications' },
+          { label: 'Resources', value: 'resoureces' },
```
sanelo: Typo on resources here

packages/breadcrumbs/Breadcrumbs.js
GoPro16: Right, it can go both ways but the idea was that you should be submitting feedback generally about the app and not the "page". So if you had a feedback component on the Home page it doesn't make sense to display `Tell us what you think about Home`. Although this can also be overridden in the props for the header text of the feedback.

sanelo: To keep the data consistent, appName should always be the same.

GoPro16: Agreed with @sanelo. The appName will be sent in the Splunk analytics and we should just require the appName on the feedback props to ensure the right analytics get logged.

PR #70, opened 2019-03-05
Pagination/addons

GoPro16:
[Inspiration](https://github.com/AdeleD/react-paginate)

They have a similar library here but there would have to be a significant PR there in order to get the desired results we want with hooks. I plan on looking into this in the future but for now I have added similar logic.

Features:
- onPageChange callback for `Pagination` Provider
- Margin Pages & Page Ranges for `PaginationControls`

screen shot 2019-03-05 at 10 47 20 am

### DIFF
--- a/packages/pagination/src/PaginationControls.js
+++ b/packages/pagination/src/PaginationControls.js
@@ -45,7 +45,7 @@ const PaginationControls = ({
   const createBreak = index => (
     <PaginationItem key={index} data-testid={`control-page-${index}`}>
       <PaginationLink onClick={() => handleBreakClick(index)} type="button">
-        ...
+        &hellip;
       </PaginationLink>
     </PaginationItem>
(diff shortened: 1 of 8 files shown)


packages/pagination/src/PaginationControls.js
```diff
@@ -1,10 +1,98 @@
...
+  };
+
+  const handleBreakClick = index => {
+    setPage(currentPage < index ? getForwardJump() : getBackwardJump());
+  };
+
+  const createBreak = index => (
+    <PaginationItem key={index} data-testid={`control-page-${index}`}>
+      <PaginationLink onClick={() => handleBreakClick(index)} type="button">
+        ...
```
TheSharpieOne: use `&hellip;`

robmcguinness:
👏

PR #71, opened 2019-03-06
feat(docs): upgrading storybook to v5

GoPro16:
Storybook v5 was just released and there are some other addons that need to be either updated or replaced before we can finish this.

The `storybook-readme` addon is still on v4 and there is an open PR to get it upgarded:
https://github.com/tuchk4/storybook-readme/issues/118

We could use `addon-info` but it doesn't display as a decorator on the righthand side as well as the display port for the markdown is not that great.

Going to keep this in the draft state until this is ready.
screen shot 2019-03-06 at 9 19 24 am


robmcguinness:
`storybook-readme@5.0.0-alpha.159` released with 5.0 support https://github.com/tuchk4/storybook-readme/issues/118#issuecomment-472913144

GoPro16:
I had to add the `StoryPreview` prop to the `readme` param object because it will default to adding an inline storybook style. See https://github.com/tuchk4/storybook-readme/issues/132

PR #73, opened 2019-03-06
feat(pagination): adds watchlist and context

GoPro16:
If you are using a function to get items within the `items` prop the only way the Pagination component knows to re-render is when the `currentPage` changes or some other prop that is present in the pagination provider. You can now pass additional `watchList` params to the PaginationProvider that will essentially add to the `componentShouldUpdate` List [here](https://github.com/Availity/availity-react/blob/pagination/exports/packages/pagination/src/Pagination.js#L87).

The Context for the Pagination Component is also exported for uses in class based components.


packages/pagination/types/Pagination.d.ts
```diff
@@ -3,6 +3,7 @@ export interface PaginationProps {
     itemsPerPage?: number;
     children?: React.ReactType;
     onPageChange?: Function;
+    watchList?: Array<Object>;
```
TheSharpieOne: Seems like this should allow an array of anything, not just array of objects.

PR #76, opened 2019-03-13
fix(pagination): fixed range error when records go over a certain limit

GoPro16:
BREAKING CHANGE: pagination provider returns page count rather than array of pages

### DIFF
--- a/packages/pagination/src/PaginationControls.js
+++ b/packages/pagination/src/PaginationControls.js
@@ -12,8 +12,7 @@ const PaginationControls = ({
   marginPages,
   ...rest
 }) => {
-  const { pages, currentPage, setPage } = usePagination();
-  const pageCount = pages.length;
+  const { pageCount, currentPage, setPage } = usePagination();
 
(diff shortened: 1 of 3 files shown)


packages/pagination/src/PaginationControls.js
```diff
@@ -73,30 +72,29 @@ const PaginationControls = ({
       }
 
       let breakView;
-
-      pages.forEach((pageNumber, index) => {
+      let index;
+      let pageNumber;
+      for (index = 0; index < pageCount; index++) {
```
sanelo, line 77: Since `index` is only used in this scope, I would define it in the `for statement`, ie `let index = 0`, instead of above.

GoPro16: updated in #78 

https://github.com/Availity/availity-react/pull/78/commits/ea3d0e4331c813d8c85608ce50533ac58f42fd18#diff-912d664f4d49898b10ffaac011b8cb5eR78

PR #81, opened 2019-03-18
feat(favorite-heart): adds favorite heart package

### LINKED ISSUE #80: add <AvFavoriteHeart /> component
It would be nice if we had an open source, react `<AvFavoriteHeart />` component that mimics our internal `<av-favorite-heart />` angular directive built by @KaseyPowers. 

I will build it but opening issue here to collect ideas anyone has.

GoPro16:
Favorite Heart Icon

Todos -
- [x] unit tests
- [x] find a better way to add styling

closes #80


GoPro16:
Still adding the above todos but wanted to get some early feedback as well while I am finishing those up.

GoPro16:
@KaseyPowers  @robmcguinness  I looked around for other libraries that are doing a "favoriting" component and this one is much smoother with the styles than the alternatives I have found. We would be maintaining this `scss` but its vary minimal. 

I removed the import inside of the component in case there is a time where someone just wants to use the custom hook provided as `useFavorites`.

KaseyPowers:
Are you planning on adding listeners for AvMessage in the context? 
So that a favorite used in a body app could update when favorites are modified by the navbar.

packages/favorite-heart/FavoriteContext.js
```diff
@@ -0,0 +1,150 @@
...
+}
+
+const FavoritesContext = createContext(new Favorites());
+
+export const useFavorites = id => {
+  const { favorites, deleteFavorite, addFavorite } = useContext(
+    FavoritesContext
+  );
+
+  const [isFavorited, toggleFavorited] = useToggle(
```
KaseyPowers: would isFavorited correctly update here? from my understanding of this, the initial state is only run initially?

GoPro16:
> Are you planning on adding listeners for AvMessage in the context?
> So that a favorite used in a body app could update when favorites are modified by the navbar.

I actually have not implemented that. Will add that in.

packages/favorite-heart/FavoriteContext.js
GoPro16: Yes because that first time is just for initial state and any state update after that uses the `toggleFavorited` method which updates the local state and the global list of favorites.

GoPro16:
Alright the last thing that I have to do is mock the `avMessages` which is tough given that we do a bunch of logical checks to make sure they are on the `availity` domain. We could just abstract this more to allow it to be used in other people's projects other than availity.

packages/favorites/FavoritesContext.js
```diff
@@ -0,0 +1,166 @@
...
+    const result = await avSettingsApi.getApplication(NAV_APP_ID);
+
+    setFavorites(get(result, 'data.settings[0].favorites'));
+  };
+
+  useEffectAsync(async () => {
+    avMessages.subscribe(
+      AV_INTERNAL_GLOBALS.FAVORITES_CHANGED,
+      (event, data) => {
+        console.log('event just occured', data, 'event:', event);
```
nylon22: do we want to leave this in?

GoPro16: Lol Nope. Luckily it wasn't something like this:
http://www.commitlogsfromlastnight.com/

nylon22: 🤣

PR #82, opened 2019-03-18
chore(docs): trying to increase performance on the doc builds

GoPro16:
Looks like turning the debug logs for the build to off decreased the build time by 30 seconds and also removing the ts loaders helped some as well.


GoPro16:
@robmcguinness  can you clear the cache on netlify for this? I think it should install and build faster to success this time. Or get really close.

robmcguinness:
seemed faster.

PR #83, opened 2019-03-22
feat(app-icon): added image src and alt properties to app-icon

GoPro16:
In some cases users can use a logo or image for the app icon and still maintain the proper aspect ratios. Added in `src` and `alt` props if you want to display the `AppIcon` as an image.


robmcguinness:
LGTM

PR #86, opened 2019-04-01
Feat/async select

### LINKED ISSUE #84: Consider switching async dropdown components
Opening to measure reception to switching dropdown components.

From this: https://github.com/TheSharpieOne/react-select
To this: https://github.com/vtaits/react-select-async-paginate

The current one we're using is a fork of the original component, intended to be merged back into the original project. However, the maintainer of react-select has not responded to getting it merged in. As a result, the fork has fallen behind the original project. 

react-select-async-paginate is on top of the original component - and as a result doesn't have the same issues of being behind. 

It would probably be safest to cut a major release if we switched since the prop names are different. We could try mapping them all but it would be easy to make a mistake.

nylon22:
closes #84

### DIFF
--- a/packages/reactstrap-validation-select/AvResourceSelect.js
+++ b/packages/reactstrap-validation-select/AvResourceSelect.js
@@ -115,15 +115,13 @@ class AvResourceSelect extends Component {
     const Tag = this.props.label ? AvSelectField : AvSelect;
     const {
       delay,
-      debounceTimeout,
+      debounceTimeout = delay,
       itemsPerPage,
       watchParams,
       cacheUniq,
       additional,
       ...rest
     } = this.props;
-    const _debounceTimeout =
-      debounceTimeout !== undefined ? debounceTimeout : delay;
     let _cacheUniq = cacheUniq;
 
     if (_cacheUniq === undefined && watchParams && this.select.current) {
@@ -140,7 +138,7 @@ class AvResourceSelect extends Component {
         loadOptions={this.loadOptions}
         pagination
         raw
-        debounceTimeout={_debounceTimeout}
+        debounceTimeout={debounceTimeout}
         cacheUniq={_cacheUniq}
         additional={{
           page: 1,
--- a/packages/reactstrap-validation-select/AvResourceSelect.js
+++ b/packages/reactstrap-validation-select/AvResourceSelect.js
@@ -13,7 +13,7 @@ class AvResourceSelect extends Component {
       getResult: PropTypes.oneOfType([PropTypes.string, PropTypes.func]),
     }).isRequired,
     getResult: PropTypes.oneOfType([PropTypes.string, PropTypes.func]),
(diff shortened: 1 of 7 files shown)


TheSharpieOne:
Probably better off starting from scratch with the new library.

nylon22:
> Probably better off starting from scratch with the new library.

Hi @TheSharpieOne , can you explain why it would be better to start from scratch? I created a canary release off this branch and have tested it out on several projects that use this component without issue.

TheSharpieOne:
I seems like it was doing stuff that it no longer needs to do thanks to the new library. Things like debounce (there is a `delay` and `debounceTimeout`) and caching (shoehorning `watchParams` into `cacheUniq`). Refactoring does a lot, but sometime you end up being stuck in the old way of thinking by just changing the old code. Starting from scratch instead of a larger refactor can help. Take it back to the drawing board, what features are really needed and don't just keep a feature or prop just because it already exists. The whole thing could probably be a function with hooks, there isn't a need for a stateful component with lifecycle methods.

packages/reactstrap-validation-select/AvResourceSelect.js
```diff
@@ -97,36 +82,71 @@ class AvResourceSelect extends Component {
...
         throw error;
       });
-  }, this.props.delay);
+  };
 
   render() {
     const Tag = this.props.label ? AvSelectField : AvSelect;
+    const {
+      delay,
+      debounceTimeout,
```
TheSharpieOne: you can remove lines 125 and 126 by using default value in this destructing assignment.

```js
const {
  delay,
  debounceTimeout = delay,
  // ...
} = this.props;
```

nylon22:
@TheSharpieOne thanks for feedback. I'll revisit the existing props/features and evaluate what is needed and what is not while also balancing that against the support cost of helping people upgrade.

I'll also switch this package to be functional components with hooks as part of this PR to keep up with the times 👍

packages/reactstrap-validation-select/AvResourceSelect.js
nylon22: d8acb777f4937bc61785c9f249dd0d43b911a9fd

good catch 👍

nylon22:
> I seems like it was doing stuff that it no longer needs to do thanks to the new library. Things like debounce (there is a `delay` and `debounceTimeout`) and caching (shoehorning `watchParams` into `cacheUniq`). Refactoring does a lot, but sometime you end up being stuck in the old way of thinking by just changing the old code. Starting from scratch instead of a larger refactor can help. Take it back to the drawing board, what features are really needed and don't just keep a feature or prop just because it already exists. The whole thing could probably be a function with hooks, there isn't a need for a stateful component with lifecycle methods.

@TheSharpieOne I decided to keep support for `watchParams` and `delay` to ease the burden of helping people upgrade. Due to time constraints, I'll create a separate issue to refactor this component into hooks (and also address some of your concerns about refactoring instead of starting from scratch). Thanks again for feedback 😄

PR #91, opened 2019-04-09
Feat/page header feedback props

nylon22:
- feat(page-header): add feedbackProps prop type
    > makes it possible to leverage all of the `<Feedback />` props inside of the `<PageHeader />` component
- feat(feedback): add staticFields prop to feedback form
    > makes it possible to send static values to feedback submission by spreading key/value pairs to `avLogMessagesApi.info` request

### DIFF
--- a/packages/docs/stories/feedback.stories.js
+++ b/packages/docs/stories/feedback.stories.js
@@ -31,6 +31,7 @@ storiesOf('Actions|Feedback', module)
               { label: 'News and Announcements', value: 'news' },
             ]
           : [],
+        staticFields: [{ name: 'staticFields1', value: 'my-static-value' }],
       }}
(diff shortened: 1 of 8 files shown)


TheSharpieOne:
This seems like a very roundabout way of getting extra data into the request.
Why not just provide an object with "static key/values" and merge/spread it into the object that gets sent to the `avLogMessagesApi.info`?

nylon22:
> This seems like a very roundabout way of getting extra data into the request.
> Why not just provide an object with "static key/values" and merge/spread it into the object that gets sent to the `avLogMessagesApi.info`?

@TheSharpieOne Yep - all i'm trying to do is get extra data into the request so that'll work too. I'll refactor 👍

PR #104, opened 2019-04-29
feat: add link component

nylon22:
A simple link component that handles the logic to make urls compatible for loadApp. Meaning, it renders a link that when clicked in a new tab, will open inside the iframe of the home page.

### DIFF
--- a/packages/docs/stories/link.stories.js
+++ b/packages/docs/stories/link.stories.js
@@ -1,0 +1,28 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import README from '@availity/link/README.md';
+import AvLink from '@availity/link';
+
+storiesOf('Components|Link', module)
+  .addParameters({
+    readme: {
+      // Show readme at the addons panel
+      sidebar: README,
+      // eslint-disable-next-line react/prop-types
+      StoryPreview: ({ children }) => <div>{children}</div>,
+    },
+  })
+  .add('with absolute url', () => (
+    <div className="py-3">
+      <AvLink url="https://github.com/Availity" target="_blank">
+        Availity Github
+      </AvLink>
+    </div>
+  ))
+  .add('with relative url', () => (
+    <div className="py-3">
+      <AvLink url="/public/apps/my-app" target="_blank">
+        My Application
+      </AvLink>
+    </div>
+  ));
--- a/packages/link/Link.js
+++ b/packages/link/Link.js
@@ -1,6 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import isAbsoluteUrl from 'is-absolute-url';
+import { isAbsoluteUrl } from '@availity/resolve-url';
 
 // if absolute, return url. otherwise loadappify the url
 export const getUrl = (url = '') => {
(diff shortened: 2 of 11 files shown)


packages/link/Link.js
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import isAbsoluteUrl from 'is-absolute-url';
+
+// if absolute, return url. otherwise loadappify the url
```
GoPro16, line 5: This package is not IE11 compatible. I would submit a PR to this repo and add the file as an export and then use it here:
https://github.com/Availity/sdk-js/blob/master/packages/resolve-url/is-absolute-url.js#L2-L8

packages/link/Link.js
```diff
@@ -0,0 +1,30 @@
...
+    {children}
+  </Tag>
+);
+
+AvLink.defaultProps = {
+  tag: 'a',
+};
+
+AvLink.propTypes = {
+  url: PropTypes.string.isRequired,
```
GoPro16, line 30: Not sure if we should or not here. But maybe still using the same default name of `href` as the prop name since it is in fact an href of the tag.

Also, I know this is bad practice but if this `Tag` ever needs to be anything other than an anchor tag say a `button`, we should expose the url as well in the `onClick` property possibly?

```javascript
// sudo code. But prop on onClick can accept the url and the event
({ onClick}) =>
onClick={(event) => onClick(getUrl(url),event)}
```

packages/link/Link.js
GoPro16: @robmcguinness  we may just want to fork that repo for `is-absolute-url` and publish to `@availity` otherwise we can just export it from the `sdk-js` package you created for `resolve-url`

nylon22: good catch: https://github.com/Availity/sdk-js/pull/105

packages/link/Link.js
nylon22: yeah - i'm fine either way calling this prop `url` or `href`. i went with `url` because what you pass in as the prop isn't necessarily what gets rendered as the `href`.  ¯\_(ツ)_/¯

i'll add the `onClick` functionality 👍

GoPro16: Its potato or potato, didn't know your thoughts on that.  We can keep as `url`

packages/link/Link.js
nylon22: 6a7d02934d8cd2432fbea67eb5dc2d575411fab3

packages/link/Link.js
nylon22: 8e3a646769d78b4ba4c6b599636bf5e2b15bda19

GoPro16:
Looks like travisci build is failing. Picking up some linter errors.

packages/docs/stories/link.stories.js
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import README from '@availity/link/README.md';
```
GoPro16, line 3: I think this is getting unresolved due to it not being in the docs `package.json`

nylon22: good catch 🤦‍♂ 00ac243390532c7ed7a7e69a463404ab433d5a48

GoPro16:
Released

PR #105, opened 2019-04-30
Fix/select error

nylon22:
fixes an issue where if the `loadOptions` call fails on an
`<AvResourceSelect />` the call gets made infinitely until the dropdown
is unfocused. By returning `hasMore: false`, we let
react-select-async-paginate know not to try to make the call again


nylon22:
> Not sure if you had not ran `npm i` before running this because travis is still expecting the previous snapshot.

yep - forget to run `npm i` again 🙃 

5569341a3fb8a70a2ba5e214bb8ded37c9c01b06

PR #110, opened 2019-05-03
fix(favorites): send post message to window.parent when favorite added

nylon22:
when `addFavorite` gets run, we send a post message to parent with the result of the settings call, letting the parent know of new favorites.

### DIFF
--- a/packages/favorites/FavoritesContext.js
+++ b/packages/favorites/FavoritesContext.js
@@ -40,7 +40,7 @@ const Favorites = ({ children }) => {
     avMessages.subscribe(
       AV_INTERNAL_GLOBALS.FAVORITES_CHANGED,
       (event, data) => {
-        setFavorites(get(data, 'favorites') || []);
+        setFavorites(get(data, 'message.favorites') || []);
       }
     );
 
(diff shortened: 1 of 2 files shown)


packages/favorites/FavoritesContext.js
```diff
@@ -40,7 +40,7 @@ const Favorites = ({ children }) => {
     avMessages.subscribe(
       AV_INTERNAL_GLOBALS.FAVORITES_CHANGED,
       (event, data) => {
-        setFavorites(get(data, 'favorites') || []);
+        setFavorites(get(data, 'message.favorites') || []);
```
nylon22, line 40: @GoPro16 the favorites that come through are inside of a `message` object

packages/favorites/__test__/FavoriteHeart.test.js
```diff
@@ -112,25 +118,6 @@ describe('FavoriteHeart', () => {
     await waitForElement(() => getByText('This item is not favorited.'));
   });
 
-  test('should call onChange once toggled', async () => {
```
nylon22, line 115: @GoPro16 i had to move this test to the end of the file to get them to pass. I don't know why. I suspect some sort of timing issue with the clean up that runs after each test.

packages/favorites/__test__/FavoriteHeart.test.js
```diff
@@ -1,6 +1,11 @@
 import React from 'react';
 import 'react-testing-library/cleanup-after-each';
```
GoPro16: We already have the `cleanup` running after each run. You shouldn't have to import the `cleanup` function as well. Since you also have to clear the mocks just remove this line as well or else its running the cleanup function twice.

nylon22: 918addb1072df98218d43a018947d24267f30254

good catch - forgot to remove when i was trying to determine why i had to move that test.

PR #118, opened 2019-05-13
feat: add spaces-image package

nylon22:
This change genericizes the existing `payer-logo` package into a `spaces-image` package, so that we can easily add `<SpacesBillboard />` and `<SpacesTile />` exports as well as any future spaces images, while reusing shared code to create them. 

As part of this change, I'm proposing:
1. Renaming `payer-logo` to `spaces-image`, since we now export more than one type of image
2. Deprecating the existing `@availity/payer-logo` package on npm
3. Switching `page-header` to use `spaces-image` instead of `payer-logo`

A few notes:
1. I added a deprecation notice to the README of the `payer-logo` package. If it would be preferred to remove it from source control entirely, please let me know.
2. I made `<SpacesLogo />` the default export of `spaces-image` to ease the burden of switching from `payer-logo` to `spaces-image`


packages/payer-image/PayerImage.js
```diff
@@ -0,0 +1,127 @@
...
+
+const PayerImage = ({ spaceId, payerId, clientId, imageType, ...props }) => {
+  const [url, setUrl] = useState(null);
+
+  useEffectAsync(async () => {
+    const _url = await getImage(spaceId, payerId, clientId, imageType);
+    setUrl(_url);
+  }, [spaceId, payerId, clientId, imageType]);
+
+  if (!clientId || (!payerId && !spaceId)) return null;
```
TheSharpieOne: I know this was copied from the existing code, but looking over it again now, shouldn't we also return null for `!url`  as well? It seems like it might hit this case now more than before since the eligibility fallback only works for logos and none of the other image types.

packages/payer-image/PayerImage.js
```diff
@@ -0,0 +1,127 @@
...
+      const path = 'spaces.spaces[0].images';
+      url = await fetchImage(
+        payerIDQuery,
+        variables,
+        path,
+        clientId,
+        imageType
+      );
+
+      // We can probably remove this at some point once our spaces data is complete
```
TheSharpieOne: It would be nice for this to be the case now and just name this package `spaces-image` and be slightly more specific to what it's going and more generic for what images it can return (not every space is a payer space 😉)

nylon22: 07d0681a52b5d6d8cda9efb584ca37d838f61aca

packages/payer-image/PayerImage.js
nylon22: 11b0b37e879680eba124e51d826c2cf355dc9cf7

packages/spaces-image/SpacesImage.js
```diff
@@ -0,0 +1,127 @@
...
+const fetchImage = async (query, variables, path, clientId, imageType) => {
+  try {
+    const {
+      data: { data },
+    } = await avSlotMachineApi.create(
+      {
+        query,
+        variables,
+      },
+      { headers: { 'X-Client-ID': clientId } }
```
GoPro16, line 40: I can see in the future where we expose some function say from `sdk-js` where we auto apply the clientId to all requests. In that case it would be nice to not require the `clientId` as a prop for some cases. If the clientId is `undefined` here, as long as it doesn't overwrite a global one that was placed before it this looks good.

Otherwise we need to add a conditional check to add that `key/value` to the request.

nylon22:
After thinking about it more, I'm declining this pr. I'm going to rewrite it to have a spaces provider that can render multiple `<SpacesImage />` all under the context of the spaces provider. It will make  this package more performant when rendering multiple spaces images by only having to make one network call rather than one per `<SpacesImage />`. Thanks @TheSharpieOne and @GoPro16 for feedback.

PR #139, opened 2019-05-30
 feat(upload): upgrade react-dropzone from 7.x to 10.x 

robmcguinness:


robmcguinness:
I think the docs should include the dropzone feature. I'd like to add before merging

PR #141, opened 2019-05-30
feat(reacstrap-validation-select): upgrade react-select 3.0

robmcguinness:
Having an issue with this upgrade. For some reason `react-select-async-paginate` default export is being imported incorrectly:

```js
import AsyncPaginate from 'react-select-async-paginate';

// ...

const Tag = attributes.loadOptions ? AsyncPaginate : Select;

```

The docs throw error when clicking on `AvSelectResource`:

> Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: undefined. You likely forgot to export your component from the file it's defined in, or you might have mixed up default and named imports. Check the render method of `AsyncPaginateBase`.

This is occurring because 'react-select-async-paginate' is being imported as:

```js
AsyncPaginateBase: (...)
default: (...)
reduceGroupedOptions: (...)
wrapMenuList: (...)
```

Somehow the default export is wrong: Downgrading resolves the issue.


GoPro16:
@robmcguinness the current version of `react-select-async-paginate` isn't compatible with `v3` of `react-select`.

Looking at the changelog for `react-select`, I think they updated a lot of the structure, as they are also moving to a monorepo and components were updated.

nylon22:
we should probably open up an issue to mak `react-select-async-paginate` compatible with `react-select@3.x`

robmcguinness:
https://github.com/vtaits/react-select-async-paginate/issues/26

robmcguinness:
need to update commit message with breaking change then we should be good.

robmcguinness:
commit message updated to include breaking change.

PR #142, opened 2019-05-30
feat(pagination): adding defaultPage prop

AdenFlorian:
For the story I'm working on, I'm replacing some legacy pagination with this newer one, and one of the requirements is that if the user clicks on an item on the page and gets taken to a different page in the same app, then goes back, it needs to be on the same page. Since the component unmounts when we show the other page it loses its state, so with this prop, we can store the current page at a higher level and pass it into the defaultPage prop when we want to show the Pagination again.

Thanks!

### DIFF
--- a/packages/pagination/src/Pagination.js
+++ b/packages/pagination/src/Pagination.js
@@ -27,7 +27,7 @@ const Pagination = ({
   watchList,
   defaultPage,
 }) => {
-  const [currentPage, setPage] = useState(defaultPage || 1);
+  const [currentPage, setPage] = useState(defaultPage);
   const [pageData, setPageData] = useState({
     total: theItems != null && !isFunction(theItems) ? theItems.totalCount : 0,
     pageCount: 0,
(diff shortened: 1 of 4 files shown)


packages/pagination/src/Pagination.js
```diff
@@ -25,8 +25,9 @@ const Pagination = ({
   onPageChange,
   children,
   watchList,
+  defaultPage,
 }) => {
-  const [currentPage, setPage] = useState(1);
+  const [currentPage, setPage] = useState(defaultPage || 1);
```
nylon22: what if `defaultPage` was explicitly set to `0` ? 

i'm thinking instead we add a default prop:

```
Pagination.defaultProps = {
    defaultPage: 1,
   ...
};
```

and then this line is:
```
const [currentPage, setPage] = useState(defaultPage);
```

adenflorian: done!

nylon22:
@AdenFlorian Can you also add `defaultPage` to [here](https://github.com/Availity/availity-react/blob/master/packages/pagination/types/Pagination.d.ts) for the ts folks?

PR #147, opened 2019-06-03
Add progress bar package

### LINKED ISSUE #146: Add progress bar package
Our progress bar styles are different and do not have equivalent props to what is provided by reactstrap so i think worthy of its own package.

Reactstrap: https://reactstrap.github.io/components/progress/

UI kit: https://availity.github.io/availity-uikit/v3/components#Progress-Bars

nylon22:
closes #146


TheSharpieOne:
Suggestion: `width` doesn't seem right as it suggest overall width of the component, maybe `fill`?

TheSharpieOne:
Also, you restrict the aria attributes to be based out of 100 (percentage). It would be better for a11y to be able to have these represent the actual progress, such as uploading 5 documents, and having 3 document completed. Visually it would look the same, but the screenreader would indicated the progress differently. You can of course default to max of 100, but should also accept and use different values, similar to [reactstrap](https://github.com/reactstrap/reactstrap/blob/master/src/Progress.js#L69-L73) ([docs](http://reactstrap.github.io/components/progress/#Max-value)).
Note: while you can override the aria attributes with the `...rest`, it's a bit of a pain since `aria-valuenow` is set to `width` and `width` is used directly as the fill percentage.

nylon22:
1dc0e11787568c2afaf642d4273f12d314031ce5 refactors this package to use `max` and `value` props (similar to reactstrap) instead of a `width` property.

PR #162, opened 2019-06-10
feat(pagination): add resetParams prop

### LINKED ISSUE #136: Pagination | ResetParams
Right now we have a `watchList` prop to watch for changes before calling an update for Pagination. There should additionally be another param call `resetParams` where they will reset the page to 1 and re-fetch or update the page data.

nylon22:
closes #136

### DIFF
--- a/packages/pagination/src/Pagination.js
+++ b/packages/pagination/src/Pagination.js
@@ -1,4 +1,4 @@
-import React, { useState, useContext } from 'react';
+import React, { useState, useContext, useEffect, useRef } from 'react';
 import PropTypes from 'prop-types';
 import isFunction from 'lodash.isfunction';
 import isEqual from 'lodash.isequal';
@@ -25,6 +25,7 @@ const Pagination = ({
   onPageChange,
   children,
   watchList,
+  resetParams,
   defaultPage,
 }) => {
   const [currentPage, setPage] = useState(defaultPage);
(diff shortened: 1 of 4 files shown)


packages/pagination/src/Pagination.js
```diff
@@ -94,6 +95,17 @@ const Pagination = ({
     }
   };
 
+  // We don't want to reset the page on the first render
+  const firstUpdate = useRef(true);
+  useEffect(() => {
+    if (firstUpdate.current) {
+      firstUpdate.current = false;
+    } else {
+      updatePage(1);
```
GoPro16, line 109: This won't work in the case the user is on the 1st page and some resetParam prop changes when the `items` prop is a function.

nylon22: can you elaborate? i'm not following 😄

GoPro16: For the below example if I am on page 1 of the `Pagination` and I update say the `id` prop then it will not force the re-render. Because its already on page 1


```jsx
<Pagination
items={() => 
 fetch(`/some-api/${id}`,{
 filters: ...someFilters
})}
watchList={[...someFilters]}
resetParams={[id]}
>
 ... Components
</Pagination>
```

nylon22: isn't that what `watchList` is for? So, for this example, `id` should be part of your `watchList` ?

nylon22: 3823a95a1bb0288049af593e9a20bfad43a3a353

GoPro16: I think we should modify the test to test for the items being rendered properly as well. That test doesn't solidify the fact the component does call the render method properly.

nylon22: f3433b5e5e52821607fcf612b2d9e5c91a094602

PR #167, opened 2019-06-13
feat(form): Form Components Using Formik

### LINKED ISSUE #96: refactor reactstrap-validation-select
See https://github.com/Availity/availity-react/pull/86#issuecomment-478800078

GoPro16:
Adds form components that are wrapped around [Formik](https://jaredpalmer.com/formik/)

Todos
- [x] Unit Tests
- [x] Typescript Defs
- [x] reactstrap-validation-select
- [x] reactstrap-validation-date
- [x] squash some bugs
- [x] Better Examples
- [x] squash all the bugs

closes #96


packages/form/src/Checkbox.js
```diff
@@ -0,0 +1,47 @@
...
+      <Input
+        name={checkValue}
+        className={classes}
+        type="checkbox"
+        {...attributes}
+        value={checkValue}
+        checked={value}
+        onChange={toggle}
+      />
+      <Label check for={checkValue}>
```
TheSharpieOne: The `for` is supposed to match the `id` of the input. Right now, clicking on the label doesn't toggle the checkbox. (same for radio)
checkbox-label

packages/form/src/CheckboxGroup.js
```diff
@@ -0,0 +1,60 @@
...
+    if (indexOfVal === -1) {
+      valueArray.push(name);
+    } else {
+      valueArray.splice(indexOfVal, 1);
+    }
+
+    setFieldValue(groupName, valueArray);
+  };
+
+  return { toggle, value: value.includes(name), ...rest };
```
GoPro16: Need to use a lodash function for this.

packages/formik-date/package.json.lerna_backup
```diff
@@ -0,0 +1,49 @@
+{
```
GoPro16: This file shouldn't be here. Not sure if you exited the `postinstall` step while it was running?

packages/formik-date/src/Date.js
```diff
@@ -0,0 +1,174 @@
...
+    const end = moment().add(max.value, max.units);
+    return day.isBefore(start) || day.isAfter(end);
+  }
+
+  return false;
+};
+
+const AvDate = ({
+  label,
+  value: valueFromProps,
```
GoPro16: I'm not sure how passing in the `value` prop is going to play with formik. Since its managing the `value` state with the metadata it could possibly have a different value than what the `value` prop of the field has.

GoPro16:
@robmcguinness  @nylon22  I don't want to over test these components as I want our tests just to cover the parts we are overriding. 

Most of the validation logic is handled by `formik`. However I do see this as an opportunity to create a sub package in `sdk-js` for yup validation method extensions and possibly just a file inside of the `form` library to have some of the fields pre-built as you can see we are doing alot of duplicate logic here such as:
```javascript
yup.object().shape({
hello: yup.string().required("This field is required.")
});
```
Can Be:
```javascript
yup.object.shape({
hello: yup.field(yup.string()).required()
});
```

Something on the above lines so we can customize the logic going into the validation and especially to make the error messages as abstracted as possible to prevent duplicates.

packages/form/src/CheckboxGroup.js
```diff
@@ -0,0 +1,62 @@
...
+export const CheckboxGroupContext = createContext();
+
+export const useCheckboxGroup = name => {
+  const { setFieldValue } = useFormikContext();
+  const { name: groupName, value = [], ...rest } = useContext(
+    CheckboxGroupContext
+  );
+
+  const toggle = () => {
+    const valueArray = [...value] || [];
```
nylon22: i think we can remove `|| []` from this line since even an empty array is truthy.

GoPro16:
Waiting on [this PR](https://github.com/jaredpalmer/formik/issues/1667) to be merged and then we can finish the `date` validation objects for `DateRange`.

We can then add the new [yup validation package](https://github.com/Availity/sdk-js/pull/125) and verify it all works.

GoPro16:
Found out the following are either going to have to be patched in their repos or polyfilled:

`yup` - `Array.from`
`formik` - `Object.assign`

Looking at [react-app-polyfill](https://github.com/facebook/create-react-app/tree/master/packages/react-app-polyfill) it looks like the `ie11.js` import has all the polyfills we need including the ones already existing in `availity-workflow`. We can finish adding 2 of those polyfills in `availity-workflow` and be sure to make a note that `@availity/form` will require  `xx.xx` version of `availity-workflow` or `xxx` polyfills.

GoPro16:
Formik Date now works as an object value. Fixed some autocomplete functionality when the date is entered in a format other than `MM/DD/YYYY`.

I believe all that is left:
- documenting the polyfills
- release `@availity/yup`
- add polyfills to `availity-workflow`

nylon22:
There are also some `<AvPatientSelect />` changes in `reactstap-validation-select` I need to carry over

PR #169, opened 2019-06-13
refactor(reactstrap-validation-date): added hide-icon prop

GoPro16:


TheSharpieOne:
How does this differ from `datepicker={false}`

Also, when you remove the icon you probably want to remove the popover as well since it is only tied to the icon (via id) and it will most likely log an error saying it cannot find the target.

GoPro16:
Because that would make too much sense 🙃 .

Going to PR and remove the `hideIcon` from `AvDate` and add the same prop name to `AvDateRange` to hide it from that view. Will also take into account the Popover.

I need to use my 👓  before making changes.

PR #170, opened 2019-06-13
fix(spaces): wait for spaces to load to default url in SpacesLogo

### LINKED ISSUE #168: <SpacesLogo /> makes call to default url before slotmachine call resolves
https://github.com/Availity/availity-react/blob/master/packages/spaces/src/SpacesImage.js#L13

The `<SpacesLogo />` renders before the call to slotmachine to fetch the space configuration resolves. Meaning an unnecessary call to the following gets made:

> /public/apps/eligibility/images/value-add-logos/<payer-id>.gif

I'm proposing we add a `loading` property to the spaces provider so that we can hold off on rendering the `img` tag until we're done loading.

nylon22:
closes #168


GoPro16:
The `loading` state will always be `true` according to the digest cycle of hooks. I would look into using [this hook](https://github.com/sw-yx/hooks/blob/master/src/index.tsx#L167-L191) for loading and possibly adding it to our `hooks` package if we don't want all 4kb of this npm  package.

robmcguinness:
would it be possible to render loading state before the image is fetched? some of our images have fixed width and height so we could pre-render a placeholder. a nice side effect of having a placeholder is that the page doesn't jump around when a bunch of images are loading.

GoPro16:
Added `react-image` as well because we need to render the loader until the image source itself has actually been downloaded from the static URL. 

Let me know if you see any issues with this. I added `skeletonProps` as the way to pass the width and height to the loader. If you guys see a better way please lets explain here. I think going forward we are going to want to get image metadata beforehand so we can render the prop width and height of the skeleton loader.

You can test the skeleton by opening the console and switching the network connection to slow 3g.

packages/spaces/src/SpacesImage.js
```diff
@@ -1,15 +1,35 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { Spinner } from 'reactstrap';
+import Skeleton from 'react-loading-skeleton';
+import Img from 'react-image';
 import { useSpace } from './Spaces';
 
+const Loader = ({ skeletonProps, ...rest }) => (
+  <span {...rest}>
+    <Skeleton width={skeletonProps.width} height={skeletonProps.height} />
```
nylon22: would we not want to spread the skeleton props?
```js 
<Skeleton {...skeletonProps}  />
```

GoPro16:
@nylon22  do we still need the below lines for the fallback image. In most cases the image never loads.

https://github.com/Availity/availity-react/blob/master/packages/spaces/src/SpacesImage.js#L11-L17

If its specific to some apps then we should make it an optional fallback prop.

nylon22:
> @nylon22 do we still need the below lines for the fallback image. In most cases the image never loads.
> 
> https://github.com/Availity/availity-react/blob/master/packages/spaces/src/SpacesImage.js#L11-L17
> 
> If its specific to some apps then we should make it an optional fallback prop.

I think it's safe to remove as long as we cut a major release for this and the page-header package

nylon22:
@GoPro16 `fallback` change looks good to me. Anything else to consider before merging this PR?

GoPro16:
Not that I can see. One issue I was running into was how to mock the image source data for testing the loading icon vs the image when it is loaded.

If we can figure that out we can fully say this component has been tested but otherwise we should be good to merge.

PR #182, opened 2019-06-24
docs(hooks): fixed react-hooks doc format

adrian-santos:
Documentation had a wrong codeblock closing line


GoPro16:
Just a heads up since the change was just for `hooks` the commit message should be `docs(hooks)` so that lerna knows its just an update to the hooks package.

adrian-santos:
Changing commit format and PR

Issue #190, opened 2019-07-03
AvDateRange Change Button Color when picker changes


adrian-santos:
Can the [Button Color](https://github.com/Availity/availity-react/blob/b9f54d419dd6a77a8b60cdde86a8efa0341aa531/packages/reactstrap-validation-date/AvDateRange.js#L392) change into Availity's Dark Blue Color when there's an input/change in [Date Range](https://github.com/Availity/availity-react/blob/master/packages/reactstrap-validation-date/AvDateRange.js#L431)?

GoPro16:
Since this is a one off case we can create `buttonProps` to the components and then you can manually update the color on the [`onChange`](https://github.com/Availity/availity-react/blob/b9f54d419dd6a77a8b60cdde86a8efa0341aa531/packages/reactstrap-validation-date/AvDateRange.js#L330) callback.

Unless there is a strong enough justification to make this a global change for the `onChange` to trigger that the above is the correct case. Feel free to submit a PR

adrian-santos:
Closing the issue. Will be using another UX pattern instead

PR #191, opened 2019-07-03
feat(reactstrap-validation-date): Adding in buttonColorOnChange props

adrian-santos:
Adding in `buttonColorOnChange props` to `AvDateRange`. This is **optional**. The default `buttonColor` is still `light`

ChangeButtonColor

 #[190](https://github.com/Availity/availity-react/issues/190)

### DIFF
--- a/packages/reactstrap-validation-date/AvDateRange.js
+++ b/packages/reactstrap-validation-date/AvDateRange.js
@@ -234,7 +234,9 @@ export default class AvDateRange extends Component {
         if (!mEnd.isBefore(mStart.add(max.value, max.units), 'day')) {
           return (
             max.errorMessage ||
-            `The end date must be within ${max.value} ${max.units} of the start date`
+            `The end date must be within ${max.value} ${
+              max.units
+            } of the start date`
           );
         }
       }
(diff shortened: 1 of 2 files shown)


GoPro16:
I'm not entirely sure on the global use case of this? If a user enters the dates into the fields rather than clicking the calendar icon that calendar icon is not going to change. If you are using it to accomplish filled in field values then I would just add this custom logic in your app and change the pR to add `buttonProps` to the `Button` to manually update the color prop.

robmcguinness:
i don't understand why this would be needed. @adrian-santos can you provide some details on what UX pattern this is solving?

adrian-santos:
@robmcguinness @GoPro16 , Our app is only using just the button since we're hiding the two date fields. There are requests from our customers that it doesn't tell them whether they already put a date in or not so they normally double check if there's a date or not. 

Right now, this works exactly the same thing as before giving the button color `light` as the default. However, when you add in `buttonColorOnChange` with a bootstrap color as the value, it changes the color of the button after they pick dates, indicating that they have already put in some dates.

It still needs some work since I only changed the `buttonColor` in the `onPickerChange`. I still need to change the `buttonColor` the user inputs both the start and end date fields.

I can change into a `buttonProps` then manually put a logic in an `onChange` function. However, not sure if this PR makes it easier when a dev wants to indicate to the user that a date has been entered. I can make an `AvDateRangeButton` component for our use case since we only need the button. We're currently hiding the `input-group-date-range> :not(:last-child)` class.

robmcguinness:
doesn't the value indicate a date was selected? I can see that in your screen capture.

adrian-santos:
Our app doesn't use the date fields to enter the data, we're only using the button.

Screen Shot 2019-07-05 at 1 52 25 PM

robmcguinness:
closing since this was resolved using different ux pattern.

PR #193, opened 2019-07-11
refactor: added styles.scss back into reactstrap-validation-select

hnicbaker:


GoPro16:
I know this isn't your fault but probably should be fixed in this PR. 

https://travis-ci.org/Availity/availity-react/builds/557476298?utm_source=github_status&utm_medium=notification

Travis build is failing because it is trying to import `lodash.isEqual` but the package name should be `lodash.isequal` lowercased if you can fix that.

PR #194, opened 2019-07-11
Fix spaces and page-header prop types

nylon22:
fix(page-header): make skeletonProps.width a string
fix(spaces): only spread skeletonProps onto Loader

resolves prop type warnings

### DIFF
--- a/packages/page-header/PageHeader.js
+++ b/packages/page-header/PageHeader.js
@@ -35,7 +35,7 @@ const PageHeader = ({
       payerId,
       className: 'float-md-right d-inline-block',
       skeletonProps: {
-        width: '180',
+        width: 180,
         height: '100%',
       },
     };
--- a/packages/page-header/PageHeader.js
+++ b/packages/page-header/PageHeader.js
@@ -35,7 +35,7 @@ const PageHeader = ({
       payerId,
       className: 'float-md-right d-inline-block',
       skeletonProps: {
-        width: 180,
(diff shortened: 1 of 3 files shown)


GoPro16:
@hnicbaker  fixed the linter issue in the travis build via #195  so we don't have to worry about that.

packages/page-header/PageHeader.js
```diff
@@ -35,7 +35,7 @@ const PageHeader = ({
       payerId,
       className: 'float-md-right d-inline-block',
       skeletonProps: {
-        width: 180,
+        width: '180',
```
GoPro16: Does changing it from number to string work? I would think you would need to add `180px` for it to still maintain the 180 pixel width.

All `react-loading-skeleton` is doing is assigning the width property to the component. [Here](https://github.com/dvtng/react-loading-skeleton/blob/master/src/skeleton.js#L53)

GoPro16: Sorry confusing you guys here, the prop-type is incorrect. It should accept `string`, `number` or `null`. The current value there is correct.
https://github.com/dvtng/react-loading-skeleton/blob/master/src/index.d.ts#L8

nylon22: Good catch 😅 

https://github.com/Availity/availity-react/pull/194/commits/800c9375746a94c1b8ca2137eb85d124ffdcd98c

PR #203, opened 2019-07-22
feat(hooks): added useTimout and useMount hooks, tests, and docs

jordan-a-young:
Added hooks `useTimeout` and `useMount`.

`useTimeout` - takes in milliseconds and returns a boolean when the elapsed time has passed.

`useMount` -  provides a hook that will run on component mount.

`useCurrentRegion` - returns the current logged in user's region`

`useDisclaimer` - given a disclaimerId will fetch and return the content of the disclaimer


packages/hooks/useTimeout.js
```diff
@@ -0,0 +1,20 @@
+import { useEffect, useState } from 'react';
+
+export default (ms = 0) => {
+  const [ready, setReady] = useState(false);
+
+  useEffect(
+    () => {
+      const timer = setTimeout(() => {
```
GoPro16: I would add a `boolean` here  called `unmounting` or something and set it to `false`. Once the timeout is up make sure to check that `unmounting` isn't set to true before you call the state update or else you will run into warnings.

packages/hooks/useTimeout.d.ts
```diff
@@ -0,0 +1,3 @@
+declare const useTimeout: Function;
```
GoPro16: this is assuming its a `const`when its a function. I also need to refactor the other ones to be correct but below is the correct way:
```typescript
declare function useTimeout(ms: number): boolean;
```

packages/hooks/useMount.d.ts
```diff
@@ -0,0 +1,3 @@
+declare const useMount: Function;
```
GoPro16: Same thing, here the typescript def should be
```typescript
/**
 * Function that will run when the component mounts
 * @returns Unmount Function that will be ran when the component unmounts
 */
type MountFunction = (...args: any) => Function;

/**
 * Runs function when the component first mounts
 * @param func Function to run when component mounts
 */
declare function useMount(func: MountFunction): void;

export default useMount;

```

packages/hooks/src/useDisclaimer.js
```diff
@@ -0,0 +1,23 @@
+import { useState, useEffect } from 'react';
```
nylon22: I wonder if we need this hook. The disclaimers api was intended to be deprecated in favor of spaces. Exposing this hook might be confusing. Instead, I think we might want to encourage people to use `@availity/spaces`. Keep in mind, `sdk-platform-disclaimers` reads from same data store as `spaces` now. Thoughts? @robmcguinness @GoPro16 

https://github.com/Availity/availity-react/blob/master/packages/spaces/src/SpacesDisclaimer.js#L9

packages/hooks/typings/useCurrentRegion.d.ts
```diff
@@ -0,0 +1,8 @@
+type CurrentRegionType = {
+    code: string;
+    value: string;
+};
+
+declare function useCurrrentRegion(): [CurrentRegionType | undefined, boolean];
```
nylon22: s/b `useCurrentRegion`

packages/hooks/typings/useCurrentRegion.d.ts
```diff
@@ -0,0 +1,8 @@
...
+
+export default useCurrrentRegion;
```
nylon22: s/b `useCurrentRegion`

packages/hooks/src/useDisclaimer.js
GoPro16: So this would work, however there are cases were we don't have the disclaimer but we have a `disclaimerId` as a piece of metadata on the field itself. So you would have two instances of `useSpace` being called.

```javascript
import { useSpace } from '@availity/spaces';
import get from 'lodash.get'

// === Function Component
const { space = {} } = useSpace(spaceId);
const { space: disclaimer = {} } = useSpace(get(space,'metadata.disclaimerId`,null)):
```

This approach is fine but we just are going to have to make a change to the `@availity/spaces` package to fetch the `space` if it doesn't exist.

Then we may have to make a decision on if this is the default behavior or if we need to start adding an `options` param to the `useSpace` or `SpacesProvider` for `fetchingIfNotExist`. We would also have to keep track of the ids that we passed in the `SpacesProvider` props so that we don't refetch on ones we have already tried.

nylon22: I don't think the tradeoff in complexity is worth it. Why not just require that the consumer of the spaces package pass in `space.metadata.disclaimerId` in the `spaceIds` prop of the `SpacesProvider` ?

We already fetch any new spaces that get passed in: https://github.com/Availity/availity-react/blob/master/packages/spaces/src/Spaces.js#L125, so if the consumer needs the initial space first to get the `space.metadata.disclaimerId`, we support that use case.

GoPro16: Not sure if I am following completely here. Are you saying to just have another `SpacesProvider` tag nested that will be the one that renders the disclaimer? Putting the newly fetched `disclaimerId` back in the root `SpacesProvider` would assume there is some prop passed down for updating the list of `spaceIds` and `payerIds` that was originally passed into that root provider.

GoPro16: Deleted `useDisclaimer` and added `SpacesAgreement` component.

PR #208, opened 2019-07-25
feat(reactstrap-validation-select): add AvPatientSelect

nylon22:
Also adds support for graphql apis in `AvResourceSelect`. 

GraphQL support in `AvResourceSelect` assumes the GraphQL api follows the relay spec.

Also adds `autofill` props to autofill corresponding values for a select dropdown in the form


packages/reactstrap-validation-select/resources.js
```diff
@@ -46,6 +47,46 @@ const AvCodeSelect = AvResourceSelect.create({
...
+    `${option.lastName}, ${option.firstName} (${option.memberId ||
+      option.subscriberMemberId})`,
+  requiredParams: ['customerId'],
+  watchParams: ['customerId', 'providerUserId'],
+  hasMore: false, // pagination not supported
+  getResult: data => data.data.patientsMany,
+  graphqlConfig: {
+    type: 'patient',
+    query: `query($filters: PatientFilters) {
+  patientsMany(filters: $filters) {
```
GoPro16, line 70: Shouldn't we be using the `patientsPagination` query here?

nylon22: The `coverages` api, which is what `thanos` is pulling patient data from (for now) - does not support pagination. So the `patients` query in `thanos` doesn't conform to the spec of other aries 1 api queries.

GoPro16: Ah sounds good, I am going to remove `thanos` from `sdk-js` and rename it differently as its going to replace all of those endpoints eventually. The name is going to change as well since copyright.

packages/reactstrap-validation-select/resources.js
```diff
@@ -47,6 +50,153 @@ const AvCodeSelect = AvResourceSelect.create({
   watchParams: ['list'],
 });
 
+/* eslint-disable react/prop-types */
+const PatientOption = props => {
```
GoPro16: Can we move this to another file? Just to the `resources.js` file clean.

nylon22: yeah good call. i'll update

nylon22:
> Looks legit! Also need to add this change the to formik components as well.

I'll add to `feat/form`

packages/reactstrap-validation-select/resources.js
nylon22: 7058a9931b8c459c0efacc28872e087864893352

nylon22:
> Not sure if you want to wait until we release in `sdk-js` package with some equivalent of `thanos`

yeah it can wait

TheSharpieOne:
You should allow autofill to be an object, if it is an object, it is a map of the field to input name. I did something similar with the Express Entry dropdown fields for the apps I built (you may be able to use it for reference).

It would also allow for nesting. 

```
{
  "payer": {
    "name: "Availity",
    "id": 1
  }
}
```
could be mapped to input names like "payerName" and "payerId" given a map like:
```
{
  "payerName": "payer.name",
  "payerId": "payer.id"
}
```
You basically just use the keys to determine the input name and use lodash's `get` with the value on the object to get the value to populate.

packages/reactstrap-validation-select/AvSelect.js
```diff
@@ -171,6 +172,53 @@ class AvSelect extends AvBaseInput {
...
+            !!inputValue.label &&
+            !!inputValue.value &&
+            typeof inputValue.value === 'object'
+          ) {
+            rawValue = inputValue.value;
+          }
+
+          let inputValueIsInForm = false;
+          if (typeof this.props.autofill === 'object') {
+            inputValueIsInForm = Object.keys(this.props.autofill).some(
```
TheSharpieOne: Not 100% sure what this is doing. Wouldn't something like `this.props.autofill[fieldName]` determine the same thing?
If would filter things that wouldn't be too useful in the `get` below. And if you want to keep those things, `Object.prototype.hasOwnProperty.call(this.props.autofill, fieldName)` would also work (or just lodash's `has`).

nylon22: The purpose of that is to check if that form input has a corresponding key in `autofill` (if `autofill` is an object) or in the `rawValue` (if `autofill` is a bool). If it does, we update the form input with the autofilled value. Maybe `inputValueIsInForm` would be better named `autofillKeyIsInForm`. 

`this.props.autofill[fieldName]` would also work - but not for the `else` statement. we would still want to autofill the component if `rawValue[fieldName]` is falsy. I'll update.

I don't understand what you mean in this part:

> If would filter things that wouldn't be too useful in the get below. And if you want to keep those things, Object.prototype.hasOwnProperty.call(this.props.autofill, fieldName) would also work (or just lodash's has).

nylon22: Please see e34cac45ed3ee41d79b6effafb50767b4e7abd97 . Thanks for feedback!

TheSharpieOne: For that other part, I mean if `this.props.autofill[fieldName]` is `false`, the `.some` would still return `true` since the key exists, but the value `false` in the `get` wouldn't really work...

nylon22: Makes sense now. Added 4740d97b56dd63865ea391f8cda992e714993118 Thanks again! 🙌

PR #212, opened 2019-07-29
docs: updated AvResourceSelect payload docs

jeffski13:


GoPro16:
LGTM.

As a side note its not going to bump all the projects since the subject is `docs` however if since its part of the `reactstrap-validation-select` package the `scope` should be included so it would be something like `docs(reactstrap-validation-select): comment`

I am going to update the [CONTRIBUTING.md](https://github.com/Availity/availity-react/blob/master/CONTRIBUTING.md) later with some examples to better convey this.

PR #215, opened 2019-07-30
Feat/react dates

GoPro16:
Replaced `react-date-range` with `react-dates`.

`reactstrap-validation` is definitely in need of some love with new the context api to help clean up this code some. For now it's working with 2 hidden input fields to store the values rather than registering as one field with an object.

BREAKING CHANGE: new look and feel as well as well as removing some validation methods

Todos
- [x] Update README
- [x] Unit Tests
- [x] Update Types


GoPro16:
Added some simple unit tests. There are more added in the #167  that will help with this when we move over.

packages/reactstrap-validation-date/package.json
```diff
@@ -36,9 +37,11 @@
   },
   "dependencies": {
     "@availity/icon": "^0.4.2",
-    "dayjs": "^1.8.14",
+    "classnames": "^2.2.6",
+    "moment": "^2.24.0",
     "prop-types": "^15.5.8",
-    "react-date-range": "^0.9.4"
+    "react-dates": "^20.2.5",
+    "react-with-direction": "^1.3.0"
```
nylon22, line 44: I see this dependency got added but don't see where it's used anywhere. do we need it?

packages/reactstrap-validation-date/src/AvDateRange.js
```diff
@@ -0,0 +1,357 @@
...
+          <DateRangePicker
+            disabled={attributes.disabled}
+            startDate={startDate}
+            startDateId={startId}
+            endDate={endDate}
+            endDateId={endId}
+            onDatesChange={this.onDatesChange}
+            focusedInput={focusedInput}
+            onFocusChange={this.onFocusChange}
+            isOutsideRange={isOutsideRange(min, max)}
```
nylon22: do we need the `min` and `max` props for this? or can we use the `distance` prop or `start` and `end` props to derive it? If we do need it, I think we should add to README. Same for `AvDate`

packages/reactstrap-validation-date/src/AvDate.js
```diff
@@ -0,0 +1,217 @@
...
+    const hasError = this.context.FormCtrl.hasError(name);
+
+    const classes = classNames(
+      className,
+      touched ? 'is-touched' : 'is-untouched',
+      this.context.FormCtrl.isDirty(name) ? 'is-dirty' : 'is-pristine',
+      this.context.FormCtrl.isBad(name) ? 'is-bad-input' : null,
+      hasError ? 'av-invalid' : 'av-valid',
+      touched && hasError && 'is-invalid',
+      !this.state.value && 'current-day-hightlight'
```
nylon22: i know that's how it's defined in css file but i'm concerned someone sees that as a typo and ~fixes~ breaks it

packages/reactstrap-validation-date/styles.scss
```diff
@@ -1,3 +1,150 @@
-.popover.popover-calendar, .popover.popover-calendar-range {
```
nylon22, line 1: i believe border radius on the input needs to be changed to look like a reactstrap input.

packages/reactstrap-validation-date/package.json
GoPro16: `react-with-direction` is a `peerDependency` for `react-dates`

packages/reactstrap-validation-date/src/AvDate.js
GoPro16: Yeah that needs to be fixed. typo. If there's a better way for optional highlighting the today's date I'm for it.

packages/reactstrap-validation-date/styles.scss
GoPro16: Yeah the selector's for `first-of-type` and `last-of-type` are not working properly. Updating them now.

packages/reactstrap-validation-date/src/AvDateRange.js
GoPro16: Weird I think we never actually documented that you can input a `min` or `max` date in the component itself. Since that is something built into the `AvInput` field. 

We should probably do some logic to where we check if the `validate` prop has `max` and if not default to using the `max` from the standard props if that makes sense. I am seeing that if someone wants to limit the dates allowed in the datepicker but also want a custom error message they have to define it twice.

GoPro16:
Not sure what we want to do when the user says they want to hide the calendar icon. Do we also want to hide the datepicker itself as well? I think its definitely something nice for people with screen-readers and disabilities.

nylon22:
> Not sure what we want to do when the user says they want to hide the calendar icon. Do we also want to hide the datepicker itself as well? I think its definitely something nice for people with screen-readers and disabilities.

I think it's good as is. Don't use the datepicker component if you don't want a datepicker imo

PR #216, opened 2019-07-30
Feat/block ui

GoPro16:
Moves over `react-block-ui` from [here](https://github.com/Availity/react-block-ui).

Getting this added here since its not as large and easier to maintain in one place. Storybook also provides a better way for users to interact and see updates as we make them to our components we use the most.

Todos
- [ ] Redux Block UI
- [ ] Typescript Definitions
- [ ] Unit Tests
- [ ] Storybook Docs
- [ ] README


TheSharpieOne:
Can we keep the package name the same?

GoPro16:
Yeah should still be fine. Only concern is that now its going to be non-compiled `es6` so will have to make a strong note in this package about. Unless we want to compile this one down since we know its used quite a lot by more developers.

packages/block-ui/src/BlockUI.js
```diff
@@ -0,0 +1,185 @@
...
+  ...attributes
+}) => {
+  const [top, setTop] = useState('50%');
+  const helper = useRef();
+  const blocker = useRef();
+  const topFocus = useRef();
+  const container = useRef();
+  const messageContainer = useRef();
+
+  const keepInViewFunc = () => {
```
TheSharpieOne: I think because this gets defined every render, it will cause a memory leak when attaching and attempting to detach the listener. My thought is that this function is created and then attached by reference. Later, when attempting to detach, `keepInViewFunc` is no longer the same reference as the one that was attached and thus the `window.removeEventListener('scroll', keepInViewFunc);` doesn't find an existing listen with the new `keepInViewFunc`.

`useRef` may be able to solve this as the ref(erence) can be a function (or any referencable thing (such as an object)) and not just a react ref.

TheSharpieOne:
> Yeah should still be fine. Only concern is that now its going to be non-compiled `es6` so will have to make a strong note in this package about. Unless we want to compile this one down since we know its used quite a lot by more developers.

I would like this transition to be transparent so that no one even knows it happened. Such as not needing to install a different package (or try to determine which package to use) and not needing to adjust workflows (keeping the compiled dist in the npm package). Moving the repo was for maintainability and shouldn't effect usage.

GoPro16:
Makes sense, I did re-write it with hooks which will require a breaking change bump anyway. I was thinking once we get the new form stuff released to make `availity-reactstrap-validation` its own mono-repo and move the `reactstrap-validation-date` and `reactstrap-validation-select` over to that repo.

packages/block-ui/src/BlockUI.js
```diff
@@ -0,0 +1,185 @@
...
+      };
+    }
+  }, [blocking]);
+
+  useEffect(() => {
+    if (keepInView) {
+      attachListeners();
+      return detachListeners;
+    }
+  }, [keepInView, blocking, attachListeners, detachListeners]);
```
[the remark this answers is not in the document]
GoPro16: Yeah I should have just placed them inside the effect that runs for the `keepInView`. I was refactoring and ended up leaving this out.

packages/block-ui/src/BlockUI.js
GoPro16: I think I am just going to put the declaration of the function inside of the effect.

I could possibly just use the `useCallback` hook that will create a memoized function.

GoPro16:
Closing this, as I think we should first just clean up this project in `react-block-ui` and if/when we are ready to compile all of our dependencies we can think about moving it over to here.

Ideal path:
- refactor to hooks
- rewrite unit tests in react-testing-library
- update docs to use gatsby ( possibly create `gatsby-theme-reactstrap` for possible use in other single component repos
- use microbundle to compile deps or rewrite to typescript to use `tsdx`

PR #219, opened 2019-08-02
refactor(reactstrap-validation-select): update patient option styling

nylon22:
Styling only change to patient dropdown options per ux feedback. No
functionality changes

### DIFF
--- a/packages/reactstrap-validation-select/resource-components.js
+++ b/packages/reactstrap-validation-select/resource-components.js
@@ -1,6 +1,7 @@
 /* eslint-disable react/prop-types */
 import React, { Fragment } from 'react';
 import { components } from 'react-select';
+import { Row, Col } from 'reactstrap';
 
 export const PatientOption = props => {
   const {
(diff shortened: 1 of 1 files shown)


packages/reactstrap-validation-select/resource-components.js
```diff
@@ -15,34 +16,35 @@ export const PatientOption = props => {
 
   const option = (
     <Fragment>
-      <strong>
-        {lastName}, {firstName} {middleName ? `${middleName}.` : ''}
-      </strong>
-      &nbsp;&nbsp;&nbsp;{payerName} {memberId || subscriberMemberId}
+      <Row>
```
GoPro16: I believe adding `noGutters` prop to the `Row` tag will make sure there is no horizontal scroll bar on the component unless the content actually exceeds the max length of the container.

nylon22: a687af1d4bf939e6d11fd367786c085e8024d417

PR #221, opened 2019-08-05
refactor(reactstrap-validation-select): show lastName as patient value

nylon22:
this is to handle the "creatable" use case in `<AvPatientSelect />`. We may not always have all the patient information to show as the value in the dropdown as it is currently designed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -88,7 +88,14 @@
     "extends": [
       "@commitlint/config-conventional",
       "@commitlint/config-lerna-scopes"
-    ]
+    ],
+    "rules": {
+      "header-max-length": [
+        0,
+        "always",
+        100
+      ]
+    }
   },
   "prettier": {
     "singleQuote": true,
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -88,7 +88,14 @@
     "extends": [
       "@commitlint/config-conventional",
       "@commitlint/config-lerna-scopes"
-    ]
+    ],
+    "rules": {
+      "header-max-length": [
+        0,
+        "always",
+        100
```
GoPro16, line 96: 🙏 ⬆️

packages/docs/stories/mocks/AvFormResults.js
```diff
@@ -18,6 +18,7 @@ class AvFormResult extends Component {
         <AvForm
           {...this.props}
           onSubmit={this.onSubmit}
+          /* eslint-disable-next-line no-console */
           onInvalidSubmit={() => console.log('invalid')}
```
TheSharpieOne, line 22: You could use storybook actions

PR #222, opened 2019-08-07
feat(avatar): add user avatar component

### LINKED ISSUE #197: Create User Avatar component
Should render user's avatar from `/settings` - defaulting to `/public/apps/my-profile/images/Avatars-00.png` if the user doesn't have one saved in `/settings`.

nylon22:
closes #197


packages/avatar/package.json
```diff
@@ -0,0 +1,39 @@
...
+  "publishConfig": {
+    "access": "public"
+  },
+  "devDependencies": {
+    "@availity/api-axios": "^5.4.0",
+    "@availity/api-core": "^5.4.0",
+    "@availity/localstorage-core": "^2.6.1",
+    "prop-types": "^15.5.8",
+    "react": "^16.8.3",
+    "react-dom": "^16.8.3"
```
GoPro16, line 30: From that failed test looks like `axios` is missing from `devDeps`

packages/avatar/package.json
```diff
@@ -0,0 +1,39 @@
...
+  },
+  "peerDependencies": {
+    "react": "^16.3.0"
```
GoPro16, line 37: Should probably add peerDeps like the hooks package: [Here](https://github.com/Availity/availity-react/blob/master/packages/hooks/package.json#L34-L37)

packages/avatar/Avatar.js
```diff
@@ -0,0 +1,86 @@
+import React, { useState } from 'react';
+import PropTypes from 'prop-types';
+import Skeleton from 'react-loading-skeleton';
+import Img from 'react-image';
+import { useEffectAsync } from '@availity/hooks';
+import { avSettingsApi } from '@availity/api-axios';
+import get from 'lodash.get';
```
GoPro16, line 7: I don't think this was included in the deps.

packages/avatar/Avatar.js
```diff
@@ -0,0 +1,86 @@
...
+    return (
+      <Loader
+        data-testid="avatar-img-loader"
+        skeletonProps={skeletonProps}
+        {...props}
+      />
+    );
+  }
+
+  if (!url && !loading && fallback) {
```
GoPro16: Looks like you can just add the `fallback` as the default for the `lodash.get` when it fetches the avatar.

nylon22: hmm... what about on first render before `avSettingsApi.getApplication` resolves?

GoPro16: You already have the initial `loading` set to `true` so just set the `initialAvatar` to the fallback.

packages/avatar/package.json
nylon22: ec9883842bfdb01dc396cae2bde3c9d2f36dc4bb

packages/avatar/package.json
nylon22: ec9883842bfdb01dc396cae2bde3c9d2f36dc4bb

packages/avatar/Avatar.js
nylon22: ec9883842bfdb01dc396cae2bde3c9d2f36dc4bb

packages/avatar/Avatar.js
nylon22: ec9883842bfdb01dc396cae2bde3c9d2f36dc4bb

GoPro16: You can just put the fallback here as the initial value of the avatar, and then also in the `get` call you can set the default if it doesn't get it.
https://github.com/Availity/availity-react/commit/ec9883842bfdb01dc396cae2bde3c9d2f36dc4bb#diff-ab6fd2a8dae5b527e60a017e05667e9eR31

nylon22: 39c28b4297891be663a925bf79c2c0c8456d3e4b

packages/docs/stories/avatar.stories.js
```diff
@@ -0,0 +1,25 @@
...
+    readme: {
+      // Show readme at the addons panel
+      sidebar: README,
+      // eslint-disable-next-line react/prop-types
+      StoryPreview: ({ children }) => <div>{children}</div>,
+    },
+  })
+  .add('default', () => (
+    <div className="py-3">
+      <Avatar />
```
GoPro16: Since the images aren't going to load when this is deployed we should add a fallback text knob and set the default to something like [placeholder image](https://upload.wikimedia.org/wikipedia/commons/7/7c/Profile_avatar_placeholder_large.png)

And then also adding a default width just for the example so the loader doesn't span the whole width of the screen

nylon22: 997309a421cba24ea8434ef238a26c2f3499c109

PR #225, opened 2019-08-09
feat(hooks): added useCurrentUser hook

jordan-a-young:
This hook fetches the currently logged in user

### DIFF
--- a/packages/hooks/README.md
+++ b/packages/hooks/README.md
@@ -155,9 +155,29 @@ import React, { useState } from 'react';
 import { useCurrentRegion } from '@availity/hooks';
 // ...
 const Component = () => {
-  const [region,loading] = useCurrentRegion();
+  const [region, loading] = useCurrentRegion();
 
-  return <div>Test Component</div>;
+  return <div>{loading ? 'Loading...' : region.value}</div>;
+};
+// ...
+```
+
+
+#### useCurrentUser
+
+Hook that will return the current user's region.
+
+
+##### useCurrentUser Usage
+
+```jsx
+import React, { useState } from 'react';
+import { useCurrentUser } from '@availity/hooks';
+// ...
+const Component = () => {
+  const [user = {}, loading] = useCurrentUser();
+
+  return <div>{loading ? 'Loading...' : user.id}</div>;
 };
 // ...
 ```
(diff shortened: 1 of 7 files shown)


robmcguinness:
curious: what would be the value in this over just calling the sdk directly?

packages/hooks/src/useCurrentUser.js
```diff
@@ -0,0 +1,31 @@
+import { useEffect, useState } from 'react';
```
nylon22, line 1: Please add to README

packages/hooks/src/useCurrentUser.js
```diff
@@ -0,0 +1,31 @@
...
+  const [error, setError] = useState();
+
+  useEffect(() => {
+    let ignore = false;
+
+    const fetchUser = async () => {
+      setLoading(true);
+      try {
+        const response = await avUserApi.me();
+        if (!ignore) setUser(response);
```
nylon22, line 16: Can you explain use case for `ignore` arg?

jordan-a-young:
Hooks make it easier to manage if an async call is loading, in my opinion.
That was my main motivation here.

jordan-a-young:
This is something I copied from Dan Abramov. The use case is, what if the
component unmounts before the API call finishes? This keeps us from calling
setState on an unmounted component.

robmcguinness:
@jordan-a-young appreciate feedback 🙌

packages/hooks/src/useCurrentUser.js
jordan-a-young: I added an example to the README as well as a story for Storybook

packages/hooks/README.md
```diff
@@ -155,9 +155,29 @@ import React, { useState } from 'react';
...
-  return <div>Test Component</div>;
+  return <div>{loading ? 'Loading...' : region.value}</div>;
+};
+// ...
+```
+
+
+#### useCurrentUser
+
+Hook that will return the current user's region.
```
nylon22: Copy pasta 😉

jordan-a-young: I thought I got em all....

GoPro16:
Going to use this PR as guinea pig for new CICD since it doesn't need to be merged immediately.

Issue #68, opened 2019-03-04
data-testid in production


TheSharpieOne:
https://twitter.com/dan_abramov/status/935661594729746434
> Add your own attributes? data-testid="...". Can then write a Babel plugin to strip them from production bundles.

Recommend using babel-plugin-jsx-remove-data-test-id or something similar to strip out the unneeded attributes from production code.

Maybe workflow should include this by default (maybe an issue should be opened there as well) but for others which just use these components outside of workflow, a note in the readme would be nice.

GoPro16:
Yeah I do like this a lot. The only downside is I am pretty sure since we are not compiling our components in here that `data-testid` will still be carried over unless people using this lib include the babel plugin as well.

I will definitely add this plugin to `@availity/workflow`.

TheSharpieOne:
Yeah, that's why it would only be a note in the readme here.

GoPro16:
Closed via [#326](https://github.com/Availity/availity-workflow/pull/326)

PR #228, opened 2019-08-13
chore: added note in readme about data-testid [skip-ci]

GoPro16:
related to #68

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ $ npm run new
 ```
 
 ## Additional Notes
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -47,6 +47,10 @@ https://availity-react-preview.netlify.com
 $ npm run new
 ```
 
+## Additional Notes
+We use `data-testid` for for unit tests to use for grabbing unique attributes. If you are using `availity-workflow@6.0.0` or later it will be removed for you in production builds. However, if you are not you will need to add the following babel plugin to remove it:
```
nylon22: "for for" s/b "for our" ?

I had trouble understanding that sentence. I think this  reads better as:
> We use `data-testid` in our unit tests to grab unique attributes...

PR #233, opened 2019-08-14
Fix/autofill default value

nylon22:
Resolves React warning that a controlled input is changing from controlled to uncontrolled by preventing field values from being set to `undefined` when auto-filling. Instead of `undefined`, sets the field value to the default/initial value

### DIFF
--- a/packages/reactstrap-validation-select/AvSelect.js
+++ b/packages/reactstrap-validation-select/AvSelect.js
@@ -209,9 +209,13 @@ class AvSelect extends AvBaseInput {
 
             let val;
             if (typeof this.props.autofill === 'object') {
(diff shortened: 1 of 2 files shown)


GoPro16:
Now that I think about it. Not sure if we want to check if there already is a value there and keep that instead of the initial value

nylon22:
> Now that I think about it. Not sure if we want to check if there already is a value there and keep that instead of the initial value

I think, no. Take a Person dropdown for example. You select "Person 1", And firstName, lastName, and birthDate get autofilled. Then you select "Person 2", but we only have firstName and lastName for person 2. You wouldn't expect the birthDate field to still have the birthDate for "Person 1".

PR #234, opened 2019-08-14
Refactor/test coverage

GoPro16:
- Adding more unit tests for test coverage
- removed unneccessary alt property check for `app-icon`.
- BREAKING CHANGE: made packages to have to install peer deps
   - `spaces`
   - `page-header`
   - `reactstrap-validation-select`
   - `favorites`
   - `payer-logo`


GoPro16:
This is good to review. Lots of updates but essentially had to restructure some of it so that they can be unit tested.

If we wanted to mock `@availity/api-axios` for a package that is using a component who relies on it, we couldn't previously because two different versions would be installed in different packages. This is now `hoisted` and also added a flag in the jest config for resolving other deps inside other packages using the `roots` attribute.

packages/page-header/README.md
```diff
@@ -7,7 +7,7 @@
 ## Installation
 
 ```bash
-npm install @availity/page-header formik yup --save
+npm install-peerdeps @availity/page-header --save
```
nylon22: s/b `npx` instead of `npm`

GoPro16: https://github.com/Availity/availity-react/pull/234/commits/98f5ce276c634f854490bcbe8c9edd03cdea12bd

PR #239, opened 2019-08-19
refactor(reactstrap-validation-date): single dateRange field filled without other should cause error

GoPro16:
Currently if we had filled out one of the date range fields but not the other it would not error out. The previous datepicker had it to where it would just put the current date of the opposite field into there and call it a day.

I don't think putting the original logic back is good for 508 as its autofilling without the user's consent and without proper reading. Instead we can just display a required message if one of the fields is filled out but not the other.


GoPro16:
oof, getting rekt by my own cicd. Going to add tests

PR #243, opened 2019-08-20
docs(form): add migration guide

### LINKED ISSUE #238: Create migration guide for availity-reactstrap-validation to @availity/form
Should provide examples of a Form using `availity-reactstrap-validation` and the equivalent in `@availity/form`

nylon22:
closes #238

### DIFF
--- a/packages/form/MIGRATION.md
+++ b/packages/form/MIGRATION.md
@@ -42,6 +42,8 @@ Another key difference is we define the initial state of the inputs in the form 
 
 ```jsx
 import { AvField, AvForm } from 'availity-reactstrap-validation';
+import '@availity/yup';
+import * as yup from 'yup';
 
 // ...
 <AvForm onValidSubmit={() => {}}>
--- a/packages/form/MIGRATION.md
+++ b/packages/form/MIGRATION.md
@@ -1,0 +1,117 @@
+# Migrating from `availity-reactstrap-validation` to `@availity/form`
+
+The primary difference in developing with `availity-reactstrap-validation` and `@availity/form` is validation. With `availity-reactstrap-validation`, we define validation schemas on the input components in the form with the `validate` prop, like so:
+
+```jsx
+import { AvInput, AvForm } from 'availity-reactstrap-validation';
+
+// ...
+<AvForm>
+  <AvInput
+    name="myInput"
+    type="text"
+    validate={{
+      required: {
+        value: true,
+        errorMessage: 'This field is required.',
+      },
+    }}
+  />
+</AvForm>
+// ...
+```
+
+With `@availity/form`, we use `yup` and `@availity/yup` to define the validation schema for all of the inputs in the form on the `<Form />` component with the `validationSchema` prop from [formik](https://jaredpalmer.com/formik/docs/guides/validation#validationschema), like so:
+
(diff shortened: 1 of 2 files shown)


packages/form/MIGRATION.md
```diff
@@ -0,0 +1,117 @@
...
+  />
+</AvForm>
+// ...
+```
+
+With `@availity/form`, we use `yup` and `@availity/yup` to define the validation schema for all of the inputs in the form on the `<Form />` component with the `validationSchema` prop from [formik](https://jaredpalmer.com/formik/docs/guides/validation#validationschema), like so:
+
+```jsx
+import { Input, Form } from '@availity/form';
+
```
GoPro16, line 44: I would show the import of `@availity/yup` and `yup`

packages/form/MIGRATION.md
```diff
@@ -0,0 +1,117 @@
...
+```jsx
+import { AvInput, AvForm } from 'availity-reactstrap-validation';
+
+// ...
+<AvForm>
+  <AvInput
+    name="myInput"
+    type="text"
+    validate={{
```
TheSharpieOne, line 13: to make this more realistic this would be:
```
<AvInput name="myInput" required />
```

packages/form/MIGRATION.md
nylon22: 97c7363636908066067542c5b3e40f01eb621d6f

packages/form/MIGRATION.md
nylon22: My goal is to bring out the comparison of the `validate` prop in `availity-reactstrap-validation` and the `validationSchema` prop in `@availity/form`. So, yes just the `required` prop is probably how most people are using the library, but I think it makes the comparison less clear. Your comment makes me think we should add a note on both the `required` prop _and_ the `validate` prop.

nylon22: d803ec1527dd3b76c881aa771a61964fac6cc1dc

packages/form/MIGRATION.md
```diff
@@ -0,0 +1,135 @@
...
+
+```jsx
+import { Input, Form } from '@availity/form';
+
+// ...
+<Form
+  initialValues={{ myInput: '' }}
+  validationSchema={yup.object().shape({ myInput: yup.string().isRequired(true, 'This field is required.') })}
+>
+  <Input name="myInput" type="text" />
```
TheSharpieOne, line 50: Should this be 
```jsx
  <Input
    name="myInput"
    type="text"
    required
  />
```
(making it kinda the same as the old way) for a11y? I'm not sure if the validationSchema is automatically added the HTML5 equivalent or not, but if not, we should show it with a11y as a11y shouldn't be an afterthought and the examples should show best practices.
It also helps keep the examples as equivalent as far as the output to the browser/user.

Similar for the ones before which involve pattern and required.

nylon22: I like the idea of automatically adding the `required` attribute to the input if it is required from `validationSchema`. I created the following issue: https://github.com/Availity/availity-react/issues/245

nylon22: I like the idea. I created #245

PR #246, opened 2019-08-23
feat(reactstrap-validation-date): added back ranges prop

GoPro16:
Adds `ranges` prop back.

- fix(reactstrap-validation-date): derived state from props was not implemented so the value prop was only valid on mount.
- refactor(icon): forwards ref to html element in the case of a ref being passed in

Trying to find a way to test those presets but no dice atm.

### DIFF
--- a/packages/reactstrap-validation-date/src/AvDateRange.js
+++ b/packages/reactstrap-validation-date/src/AvDateRange.js
@@ -17,6 +17,10 @@ import { isOutsideRange, limitPropType, isSameDay } from './utils';
 let count = 0;
 
 const relativeRanges = {
+  Today: {
+    startDate: now => now,
+    endDate: now => now,
+  },
   'Last 7 Days': {
     startDate: now => now.add(-6, 'd'),
     endDate: now => now,
@@ -24,10 +28,6 @@ const relativeRanges = {
   'Last 30 Days': {
     startDate: now => now.add(-29, 'd'),
     endDate: now => now,
-  },
-  'Last Calendar Month': {
-    startDate: now => now.startOf('month').add(-1, 'M'),
-    endDate: now => now.startOf('month').add(-1, 'd'),
   },
   'Last 120 Days': {
     startDate: now => now.add(-119, 'd'),
@@ -55,7 +55,11 @@ export default class AvDateRange extends Component {
     max: limitPropType,
     min: limitPropType,
     distance: PropTypes.object,
-    ranges: PropTypes.oneOfType([PropTypes.array, PropTypes.object]),
+    ranges: PropTypes.oneOfType([
+      PropTypes.bool,
+      PropTypes.array,
+      PropTypes.object,
+    ]),
     onPickerFocusChange: PropTypes.func,
     defaultValues: PropTypes.object,
     calendarIcon: PropTypes.node,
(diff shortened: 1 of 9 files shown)


robmcguinness:
Could you explain why this is a breaking change?

GoPro16:
Its adds a new UI element as the default that people may not want back. The ranges prop was originally omitted when we made the breaking change to v3 but now it should be another breaking change to v4 since its adding it back in.

Default: 
Screen Shot 2019-08-23 at 8 39 25 AM

robmcguinness:
Can we make it not the default and allow userland to enable the feature? I would prefer no breaking changes.

robmcguinness:
Also the first option should be `today`. I would probably git rid of `last calendar month` option but maybe others can chime in.

GoPro16:
Those ranges I put in are exactly what they were previously before we made the switch. I agree with your suggestion on removing the calendar month and adding `Today`.

I can make a change to where they are no breaking but I think we should at least include `ranges` as a boolean and if its true just include the default presets.

For some reason the netlify preview is not working properly. When running locally I am not having any issues.

robmcguinness:
>  I can make a change to where they are no breaking but I think we should at least include ranges as a boolean and if its true just include the default presets.

yes please. also it would be wild if `relativeRanges` could be passed in as well. maybe in future release

GoPro16:
> yes please. also it would be wild if `relativeRanges` could be passed in as well. maybe in future release

You can already pass in your own ranges. Its just not documented even from the previous version.

GoPro16:
@robmcguinness can you re-run the netlify build? I fixed the build issue but doesn't seem to be updating the preview.

robmcguinness:
docs deployed

packages/reactstrap-validation-date/src/AvDateRange.js
```diff
@@ -296,7 +343,10 @@ export default class AvDateRange extends Component {
       this.props.end.name
     ).getViewValue();
 
-    if (start && end) {
+    const hasStart = start && start !== '';
```
nylon22, line 346: empty strings are falsy so the second condition will never get hit if it is an empty string. so i think line 349 can just be:
```js
if (start && end)
```

packages/reactstrap-validation-date/src/AvDateRange.js
```diff
@@ -348,13 +401,84 @@ export default class AvDateRange extends Component {
...
     }
 
     return true;
   };
 
+  renderDateRanges = () => {
+    const { ranges: propsRanges } = this.props;
+    const { startValue, endValue, format } = this.state;
+
+    console.log('propsRanges', propsRanges);
```
nylon22: let's remove

packages/reactstrap-validation-date/src/AvDateRange.js
GoPro16: Hmmm, for some reason though if you didn't do this and you ran the build on the `docs`. It would actually not work properly and the validation would fail.

Steps to replicate:
- checkout ee00a6b4e04de95f3c9e377625f91dbc2b4691cd
- `npm run build:docs`
- open `index.html` file in the `docs` build folder at the root
- click a start date and the error logic will kick in incorrectly

packages/reactstrap-validation-date/src/AvDateRange.js
GoPro16: Hmmm, wondering why the linter didn't blow up on this.

https://github.com/Availity/availity-react/pull/246/commits/d86a3fd4598313595f0d82fb7048b57375962d26

packages/reactstrap-validation-date/src/AvDateRange.js
GoPro16: This is wild.
OG Function:
```js
  requireStartIfEnd = () => {
    const start = this.context.FormCtrl.getInput(
      this.props.start.name
    ).getViewValue();

    // We want the view value so not calling from args
    const end =
      (this.context.FormCtrl.getInput(this.props.end.name) &&
      this.context.FormCtrl.getInput(this.props.end.name).getViewValue());


    return (!start && end) ? 'Both start and end date are required.': true;
  };

```
Transpiled code:
```js
AvDateRange_defineProperty(AvDateRange_assertThisInitialized(_this), "requireStartIfEnd", function() {
                    var start = _this.context.FormCtrl.getInput(_this.props.start.name).getViewValue()
                      , end = _this.context.FormCtrl.getInput(_this.props.end.name) && 
                       _this.context.FormCtrl.getInput(_this.props.end.name).getViewValue();
                    return start || !end || "Both start and end date are required."
  }),
```

It returns the start date if there is one instead of the truthy value of it in that func. Transpiled code is invalid. So that workaround I put in makes it transpile correctly.

TheSharpieOne: It must assume that things which are "not"ed are booleans 🤷‍♂.

PR #250, opened 2019-09-03
fix(reactstrap-validation-select): use valueKey for newly created opt…

chrishavekost:
When the `creatable` prop is true, this fixes hardcoded `value` for cases where the `AvSelect` component is using a `valueKey`.

### DIFF
--- a/packages/reactstrap-validation-select/AvSelect.js
+++ b/packages/reactstrap-validation-select/AvSelect.js
@@ -16,9 +16,9 @@ const {
   ClearIndicator,
 } = reactSelectComponents;
 
-const createOption = (label, labelkey = 'label') => ({
-  [labelkey]: label,
-  value: label.toLowerCase().replace(/\W/g, ''),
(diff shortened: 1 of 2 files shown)


packages/reactstrap-validation-select/AvSelect.js
```diff
@@ -16,9 +16,9 @@ const {
   ClearIndicator,
 } = reactSelectComponents;
 
-const createOption = (label, labelkey = 'label') => ({
-  [labelkey]: label,
-  value: label.toLowerCase().replace(/\W/g, ''),
+const createOption = (label, labelKey = 'label', valueKey = 'value') => ({
+  [labelKey]: label,
+  [valueKey]: label.toLowerCase().replace(/\W/g, ''),
```
GoPro16, line 21: Can we add a quick test that has `creatable` and a custom `valueKey` to make sure this will work properly for future updates.

PR #269, opened 2019-09-13
style(docs): add missing perenth

VolodymyrPliuta:


GoPro16:
We need to remove links from all the summers as they don’t render properly. Instead if they are useful they should go somewhere else on the current page

VolodymyrPliuta:
good point

PR #277, opened 2019-09-16
Fix/date range same day

nylon22:
for `date` and `reactstrap-validation-date`, allows the same day to be selected as the start and the end date by leveraging the `minimumNights` prop from `react-dates`.  In `date`, it is overridable due to `datepickerProps`.


robmcguinness:
this should already be supported. @GoPro16 can you validate?

jordan-a-young:
I tried the story book example earlier today, and it was not. I have also
experienced the bug in development.

robmcguinness:
strange. one of the features request that was recently fixed was same day selection along with presets. I thought both were already solved.

robmcguinness:
i remember now: the `ranges` had the ability to select same day which is not the same as this fix. 

image

mea culpa.

robmcguinness:
i fixed the build previews. I think we are missing `minimumNights` from the knobs.

GoPro16:
> i fixed the build previews. I think we are missing `minimumNights` from the knobs.

We can also include this in the knobs or just make a note about it since its just a date picker prop that we set the default use case for.

adrian-santos:
Looks good. Thanks @nylon22

PR #280, opened 2019-09-18
Fixes Typescript Definitions For Label Props

### LINKED ISSUE #279: AvSelectField | Label Ts Def Type should be of Type String or ReactType
https://github.com/Availity/availity-react/blob/720859d6e3569d8b5ea5918616bd07a7dfe04a3d/packages/reactstrap-validation-select/AvSelectField.d.ts#L2

GoPro16:
fixes #279

### DIFF
--- a/packages/date/typings/DateField.d.ts
+++ b/packages/date/typings/DateField.d.ts
@@ -2,7 +2,7 @@ import * as React from 'react';
 import { DateProps } from './Date';
 
 export interface DateFieldProps extends DateProps {
-  label?: string;
+  label?: React.ReactNode;
 }
 
 declare class DateField extends React.Component<DateFieldProps> {}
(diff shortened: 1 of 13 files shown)


jordan-a-young:
While we are updating types, can we allow the label prop to be a string or
a react component?

Not sure if these updates do that or not as my TypeScript is weak.

GoPro16:
> While we are updating types, can we allow the label prop to be a string or a react component? Not sure if these updates do that or not as my TypeScript is weak.

Yeah its doing that. Essentially the `|` operator is an OR statement. So its `ReactType` or `String`

djejaquino:
@chrishavekost Actually `React.ReactType` is parameterized with any type, which I believe is `props` definition. `React.ElementType` should be used instead though.

GoPro16:
Correct way to do it is `React.ReactNode`.

https://codesandbox.io/s/still-cloud-c41j8

PR #285, opened 2019-09-19
Updated AvSelectFieldProps to extend AvSelectProps

### LINKED ISSUE #284: Typescript Definitions | Update reactstrap-validation-select to extend base types
Looking at `@availity/reactstrap-validation-select` we are not properly extending the base `AvSelect` Props in the `AvSelectField` and same for the reosurce select.

However, `@availity/select` is doing this correct so we can checkout how its doing it there and copy the same logic over.

djejaquino:
This is needed when working on Typescript web apps

closes #284


djejaquino:
@GoPro16 on a hurry to get this done, I force pushed amend. Hope that is ok

PR #286, opened 2019-09-20
fix(hooks): moved region fetching into useEffect, and added trycatch to handle errors

jordan-a-young:


packages/hooks/src/useCurrentRegion.js
```diff
@@ -2,26 +2,37 @@ import { useState, useEffect } from 'react';
...
+          setCurrentRegion({
+            code: response.data.regions[0].id,
+            value: response.data.regions[0].value,
+          });
+        }
+      } catch (error_) {
+        if (!ignore) setError(error_);
+      }
+
+      setLoading(false);
```
GoPro16, line 27: Just as a suggestion, we can add a `finally` block here that assures the `setLoading` happens after the `try/catch`. Otherwise LGTM

PR #287, opened 2019-09-20
chore(upload): updated peer dependency for react-dropzone

jordan-a-young:
I don't think we need all of the possible peer dependencies for `react-dropzone`, but I did not want to break any existing functionality just to fix a warning. I can update if needed.

### DIFF
--- a/packages/upload/package.json
+++ b/packages/upload/package.json
@@ -26,7 +26,7 @@
     "lodash.map": "4.6.0",
     "react": "^16.8.0",
     "react-dom": "^16.8.0",
-    "react-dropzone": "^7.0.1",
+    "react-dropzone": "^10.0.0 ||^9.0.0 ||^8.0.0 ||^7.0.1",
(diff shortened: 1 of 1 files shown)


packages/upload/package.json
```diff
@@ -26,7 +26,7 @@
     "lodash.map": "4.6.0",
     "react": "^16.8.0",
     "react-dom": "^16.8.0",
-    "react-dropzone": "^7.0.1",
+    "react-dropzone": "^10.0.0 ||^9.0.0 ||^8.0.0 ||^7.0.1",
```
GoPro16, line 29: Did `^>=7.0.1` not work?

jordan-a-young: I didn't realize I could use that syntax to be honest

GoPro16: Yeah something like: https://github.com/Availity/availity-react/blob/f9d07a08e0f4f2b5aab7762650aef29280345482/packages/authorize/package.json#L35

PR #259, opened 2019-09-06
refactor(docs): update docs to use gatsby theme WIP

nylon22:
TODOS
- [x] Squash the typos
- [x] Make sure all props match the `PropTypes` definition in the components
- [x] Get usage examples working
- [x] Standardize format
- [x] Proofread. (eg switch past/future tense to present tense for clarity/conciseness). A lot of the descriptions are difficult to read.
- [x] Update plop template for new packages to account for switch to gatsby

closes #251

PR #300, opened 2019-09-24
add infinite scroll to pagination package

### LINKED ISSUE #276: Create infinite scroll component in @availity/pagination
Should be able to reuse Pagination provider and Pagination content. Probably needs

- `infiniteScroll` prop to infinitely scroll instead of showing pagination controls. 
- `<AvResourceInfiniteScroll />` export

nylon22:
closes #276 

[Commit](https://github.com/Availity/availity-react/pull/259/commits/9ec4394e9fd3fbd8d9fe7c0da4d3681092f04410) with updated documentation in [this PR](https://github.com/Availity/availity-react/pull/259) to avoid merge conflicts

Uses https://github.com/ankeetmaini/react-infinite-scroll-component#readme


GoPro16:
Awesome component! I just added a button only available on focus to load more items if the user is using a screen reader.

However, I wonder if we can go a step further and make sure the next item is autofocused instead of focusing the load more button after being clicked.

From what I can tell the only way to do that is to ensure the `Component` prop that gets rendered in the `PaginationContent` accepts an `innerRef` prop that can be used to forward the ref to the focusable element

nylon22:
> Awesome component! I just added a button only available on focus to load more items if the user is using a screen reader.
> 
> However, I wonder if we can go a step further and make sure the next item is autofocused instead of focusing the load more button after being clicked.
> 
> From what I can tell the only way to do that is to ensure the `Component` prop that gets rendered in the `PaginationContent` accepts an `innerRef` prop that can be used to forward the ref to the focusable element

See 728da95d841757ff521b73da5bf14cfdb818641b

I believe `jsdom` has a bug that prevents us from getting a true test for this: https://github.com/testing-library/react-testing-library/issues/276

GoPro16 (approved):
> I believe `jsdom` has a bug that prevents us from getting a true test for this: [testing-library/react-testing-library#276](https://github.com/testing-library/react-testing-library/issues/276)

We can leave that comment there for FIXME and if we find an alternative or someone creates a fix we can update the test. Just from looking at the storybook preview its all working properly focusing the next element.

PR #302, opened 2019-09-26
feat(date): changed DateField and DateRangeField to allow using nodes for the label prop

stevenviar:


packages/date/src/DateField.js
```diff
@@ -6,15 +6,19 @@ import Date from './Date';
 
 const DateField = ({ name, label, ...props }) => (
   <FormGroup for={name}>
-    {label && <Label for={name}>{label}</Label>}
+    {label && typeof label === 'string' ? (
```
GoPro16: Looks like we missed some extra label attr spreading when adding in the new form library components.

The label type should look like the one linked below like reactstrap validation date:

https://github.com/Availity/availity-react/blob/master/packages/reactstrap-validation-date/src/AvDateRangeField.js#L129-L138

packages/date/src/DateRangeField.js
```diff
@@ -6,15 +6,19 @@ import DateRange from './DateRange';
 
 const DateRangeField = ({ name, label, ...props }) => (
   <FormGroup for={name}>
-    {label && <Label for={name}>{label}</Label>}
+    {label && typeof label === 'string' ? (
```
GoPro16: Same thing for this field to have the proper label attrs and children.

stevenviar:
Also, I noticed some of the tests for DateRange are failing today because it's trying to select the sibling to the current day on the calendar but it's the last day of the month.

GoPro16:
> Also, I noticed some of the tests for DateRange are failing today because it's trying to select the sibling to the current day on the calendar but it's the last day of the month.

Interesting. Maybe we can modify it to get the previous sibling if the next doesn't exist or refactor how we are just picking the date as a whole.

PR #306, opened 2019-09-30
feat(date): added additional attributes to the label for DateField and DateRangeField

stevenviar:
Creating this again because I messed the other one up rebasing...

### DIFF
--- a/packages/date/typings/DateField.d.ts
+++ b/packages/date/typings/DateField.d.ts
@@ -3,6 +3,9 @@ import { DateProps } from './Date';
 
 export interface DateFieldProps extends DateProps {
   label?: React.ReactNode;
+  labelClass?: string;
+  labelHidden?: boolean;
+  labelAttrs?: object;
 }
 
 declare class DateField extends React.Component<DateFieldProps> {}
--- a/packages/date/typings/DateField.d.ts
+++ b/packages/date/typings/DateField.d.ts
@@ -5,7 +5,7 @@ export interface DateFieldProps extends DateProps {
   label?: React.ReactNode;
   labelClass?: string;
   labelHidden?: boolean;
-  labelAttrs?: object;
+  labelAttrs?: React.HTMLAttributes<HTMLLabelElement>;
 }
 
 declare class DateField extends React.Component<DateFieldProps> {}
--- a/packages/docs/source/form/date/components/date-field.md
+++ b/packages/docs/source/form/date/components/date-field.md
@@ -52,5 +52,5 @@ The name of the class for the label. Will be passed to the `className` prop of t
 ### `labelHidden?: boolean`
 Used to control if the label is displayed. When set to `true`, the `<Label>` in the field won't be rendered.
 
-### `labelAttrs?: object`
-Pass any additional props to the `<Label>` in the field.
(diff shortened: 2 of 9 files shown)


packages/date/typings/DateField.d.ts
```diff
@@ -3,6 +3,9 @@ import { DateProps } from './Date';
 
 export interface DateFieldProps extends DateProps {
   label?: React.ReactNode;
+  labelClass?: string;
+  labelHidden?: boolean;
+  labelAttrs?: object;
 }
```
stevenviar, line 9: I see some of the old type definitions have `labelAttrs` set to `HTMLAtrributes` but do we have to restrict it like that?

GoPro16: yeah its a stronger type than object. `React.HTMLAttributes<HTMLLabelElement>`. If someone passes something invalid in if they were writing typescript react will render and it will throw an error in the console even though the typescript definition allows it.

GoPro16: In reality we should be checking if it has a custom tag and if so infer that type and use its props definition for the `labelAttrs` but this suffices

stevenviar: I was thinking that without the restriction you could pass props to the reactstrap `<Label />` that aren't necessarily attributes, but I guess we're trying to be restrictive for more control.

packages/docs/source/form/date/components/date-field.md
```diff
@@ -44,4 +44,13 @@ Extends [Date Props](/form/date/components/date/#props).
...
 ### `label?: string`
-The text that renders inside the `Label` above the input.
\ No newline at end of file
+The text that renders inside the `Label` above the input.
+
+### `labelClass?: string`
+The name of the class for the label. Will be passed to the `className` prop of the label in the field.
+
+### `labelHidden?: boolean`
+Used to control if the label is displayed. When set to `true`, the `<Label>` in the field won't be rendered.
```
TheSharpieOne: I would reword this to say the field will not be visible.
It is still rendered so that assistive technology (such as screen readers) know what the field's label is.
Same for the range docs

packages/date/typings/DateField.d.ts
TheSharpieOne: Here is what Label accepts (in addition to HTML attr (and even more if you use `tag` to render another component)):
```js
const propTypes = {
  children: PropTypes.node,
  hidden: PropTypes.bool,
  check: PropTypes.bool,
  size: PropTypes.string,
  for: PropTypes.string,
  tag: tagPropType,
  className: PropTypes.string,
  cssModule: PropTypes.object,
  xs: columnProps,
  sm: columnProps,
  md: columnProps,
  lg: columnProps,
  xl: columnProps,
  widths: PropTypes.array,
};
```

stevenviar:
Looks like you got the tests working 365 days a year, nice work

GoPro16:
Yeah sorry had done it overnight, you can check it out. Essentially I just do an if check to see what tiles should be selected and what dates should be expected. There were 2 tests that were using this logic and I removed it from the last one because it wasn't testing the clicking functionality.

PR #307, opened 2019-10-01
refactor(spaces): updated state to reducer hooks

GoPro16:
Removes 3 `useState` calls for one reducer that can handle all the scenarios.

### DIFF
--- a/packages/spaces/src/Spaces.js
+++ b/packages/spaces/src/Spaces.js
@@ -1,9 +1,4 @@
-import React, {
-  createContext,
-  useContext,
-  useMemo,
-  useReducer,
-} from 'react';
+import React, { createContext, useContext, useMemo, useReducer } from 'react';
 import PropTypes from 'prop-types';
 import { avSlotMachineApi } from '@availity/api-axios';
 import { useEffectAsync } from '@availity/hooks';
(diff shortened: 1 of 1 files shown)


packages/spaces/src/Spaces.js
```diff
@@ -56,6 +56,31 @@ export const sanitizeSpaces = spaces => {
...
+    loading: false,
+    error,
+  }),
+  LOADING: (state, { loading }) => ({
+    ...state,
+    loading: loading !== undefined ? loading : !state.loading,
+  }),
+};
+
+const spacesReducer = (state, action) => actions[action.type](state, action);
```
jordan-a-young, line 82: The only thing I can think of to be cognizant of is what will happen if `actions[action.type] === undefined`.

GoPro16: 🤷‍♂  In this case we are safe but yeah I think if it was an app built using the state reducer there would need to be an undefined check.

jordan-a-young: That's what I was thinking as well.

nylon22, line 82: would there be value in exporting this?

PR #308, opened 2019-10-01
fix(form): uses inputid attribute or generates uuid #303

### LINKED ISSUE #303: `Field` label not connected to input
### Current Behavior

`Field` is not accessible. Clicking the label does not add focus to the input.

### Expected behavior

Clicking the label generated by `Field` should give the input generated by `Field` focus.

### Reproducible example

https://availity.github.io/availity-react/form/components/field/

### Suggested solution(s)

Currently, the label's `for` attribute is set to the input's `name`. This is not how HTML works. The `for` attribute _must_ match the associated input's `id`.
Previously, `AvField` would use the `name` as the `id` if no `id` was provided. I believe a similar solution should be implemented to make the generated IDs more predictable (which is better for e2e testing tools) (or generate and cache uuid to use).

### Additional context

https://www.w3.org/WAI/tutorials/forms/labels/#associating-labels-explicitly
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label#Attributes

### Your environment

Doesn't matter.

agawish:
# What I did

- Created a local variable `inputId` from `attributes.inputId` and set it to `uuid` when it is not sat and a label is present
- Created 3 unit tests to check for the different scenarios when `inputId` exists or not and whether a label exists or not

# How I did it

1. Checked the logic for `AvSelectField` and saw that it uses `inputId` from the attributes so I followed on the same footpath
2. Followed the logic of unit tests to create my unit tests
3. `yarn test`
4. `yarn start:docs` for good measures!

# How to verify it

1. Check the unit tests in `Field.test.js`:
 - `should generate generated uuid when id is not provided`
 - `label for attribute should point to field id attribute`
 - `should generate uuid even when label is not added`
2. Run `yarn start:docs` for good measures

# A picture of a cute animal
dog-3029701_640

Fixes #303

### DIFF
--- a/packages/form/src/Field.js
+++ b/packages/form/src/Field.js
@@ -1,6 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { Input as RsInput, Label, FormText, Col } from 'reactstrap';
+import { v4 as uuid } from 'uuid';
 import Feedback from './Feedback';
 import FormGroup from './FormGroup';
 import Input from './Input';
@@ -23,6 +24,7 @@ const Field = ({
   ...attributes
 }) => {
   let row = false;
+  let { inputId } = attributes;
   const col = {};
   const labelCol = {};
 
@@ -37,9 +39,14 @@ const Field = ({
     });
   }
 
+  if (label && !inputId) {
+    inputId = uuid();
+  }
+
   const input = (
     <Input
       name={id}
+      id={inputId}
       className={inputClass}
       size={size}
       disabled={disabled}
@@ -71,7 +78,7 @@ const Field = ({
       {check && inputRow}
       {label && (
         <Label
-          for={id}
+          for={inputId}
           className={labelClass}
           hidden={labelHidden}
           size={size}
--- a/packages/form/src/Field.js
+++ b/packages/form/src/Field.js
@@ -1,7 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { Input as RsInput, Label, FormText, Col } from 'reactstrap';
-import { v4 as uuid } from 'uuid';
+import uuid from 'uuid/v4';
 import Feedback from './Feedback';
 import FormGroup from './FormGroup';
 import Input from './Input';
@@ -24,7 +24,7 @@ const Field = ({
   ...attributes
 }) => {
   let row = false;
-  let { inputId } = attributes;
+  const inputId = attributes.id || uuid();
   const col = {};
   const labelCol = {};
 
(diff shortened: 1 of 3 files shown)


nylon22 (changes requested):
i believe `uuid` s/b added as a dependency of `form` package

packages/form/src/Field.js
```diff
@@ -23,6 +24,7 @@ const Field = ({
   ...attributes
 }) => {
   let row = false;
+  let { inputId } = attributes;
```
GoPro16: I think instead of adding a new prop called `inputId` just alias or use the `attributes.id` off the field props.

Then you can declare a const that is something like the below:
`const inputId = attributes.id || uuid()`

Then make sure to set that for both the `label` and the `input` components.

GoPro16 (changes requested):

packages/form/src/Field.js
GoPro16: I believe the reason why we use `inputId` for `AvSelect` is because the `react-select` package takes in that prop and puts it on the input rather than the top level component

agawish: The reason I opted against using a `const` is that I didn't want to generate a `uuid` when label is not provided, if you're happy to generate `uuid` all the time though, I can change it!

Just for reference, the old code for AvSelect looked like this in which why I thought checking if label is needed made sense!
```
if(label){
let forLabel = id;
    if (attributes.inputId) {
      forLabel = attributes.inputId;
    } else {
      attributes.inputId = id;
    }
}
```

packages/form/src/Field.js
```diff
@@ -1,6 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { Input as RsInput, Label, FormText, Col } from 'reactstrap';
+import { v4 as uuid } from 'uuid';
```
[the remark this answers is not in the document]
agawish: Nice observation

packages/form/src/Field.js
agawish: @GoPro16 do you still want to go with your suggesstions, happy to do it?

GoPro16: Yeah you can make it conditional, my only concern was getting rid of the `inputId` prop and using the default `id` field.

agawish: I moved away from `inputId` in favour of `id`, let me know if you want to make it conditional or you're happy with the latest commit result?

GoPro16 (approved):
LGTM. We may end up removing the `uuid` in favor of the `name` prop as the default since its a required field but this is a solid addition I think and if we want to optimize later we can remove it.

Thanks for your contribution!

PR #317, opened 2019-10-03
fix(select): prevents the provided styles from react-select from bein…

stevenviar:
…g overridden by the styles applied in an error condition

Currently the custom styles that make the Select component red in an error condition are not allowing the default behavior when the select `isDisabled`.


GoPro16 (changes requested):
Can you also fix this for `@availity/reactstrap-validation-select` as well? Then we should be all good.

GoPro16 (approved):
LGTM

PR #321, opened 2019-10-04
add SpacesFile export 

nylon22:
## Description
Adds new export, `SpacesFile`, which is a pre-made `SpacesImage`.  Spaces of type `file` contain their reference to the url to be rendered under `space.url`

## Commits
- feat(spaces): add SpacesFile component
- docs(spaces): add note on variables prop

### DIFF
--- a/docs/source/components/spaces/images.md
+++ b/docs/source/components/spaces/images.md
@@ -10,6 +10,7 @@ import Spaces, {
   SpacesLogo,
   SpacesBillboard,
   SpacesTile,
+  SpacesFile,
 } from '@availity/spaces';
 
 <Spaces
@@ -20,6 +21,7 @@ import Spaces, {
   <SpacesLogo spaceId="73162546201441126239486200007187" />
   <SpacesBillboard payerId="PayerID" />
   <SpacesTile payerId="PayerID" />
+  <SpacesFile payerId="PayerID" />
 </Spaces>;
 ```
 
--- a/docs/source/components/spaces/images.md
+++ b/docs/source/components/spaces/images.md
@@ -10,7 +10,7 @@ import Spaces, {
   SpacesLogo,
   SpacesBillboard,
   SpacesTile,
-  SpacesFile,
+  SpacesImage,
 } from '@availity/spaces';
 
 <Spaces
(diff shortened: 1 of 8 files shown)


chrishavekost (approved):

GoPro16 (commented):
I am finding it a bit confusing although I am not sure what all we can do to make it more straightforward. 

`SpacesFile` could possibly be a file that is not an image IIRC. If that file is not an image then its simply a false positive that a particular `spaceId` contains an image file type.

On the other-hand, we have `SpacesImage` which takes the `spaceId` in and grabs the `imagetype` from the metadata, in most cases the `spaceId` is a payer space and not a "file".

If there is a way to make these use the same component with some additional prop or even being able to just get the `type` attribute from the space in order to know if it should be pulling the `imagetype` from metadata or the url. It just sounds weird when you are saying that you want to render a `SpacesFile` when in reality it's a `SpacesImage`.

nylon22:
@GoPro16 You're correct - type `file` could potentially not be an image. This approach assumes that consumer of this package passes the correct space id to `<SpacesFile />`, but that is true for any of the spaces components. 

I don't understand your suggestion - I think i'm more or less already doing what you're suggesting. `<SpacesFile />` is a `<SpacesImage />` with a `imageType` of `url`. See changes to `packages/spaces/index.js`. 

Slight correction - although i don't think it changes your point. The `imageType` for the other `SpacesImage` exports are inside `images`, not `metadata`.

GoPro16:
> I don't understand your suggestion - I think i'm more or less already doing what you're suggesting. `<SpacesFile />` is a `<SpacesImage />` with a `imageType` of `url`. See changes to `packages/spaces/index.js`.
> 

To an average developer. If I hear the term `SpacesFile` I would assume its a space containing a link to a file, and not a component that renders and image. I don't have a clear suggestion but rather just a comment on how this is starting to get confusing from a DX perspective.

Maybe we just remove the `isRequired` flag from `SpacesImage` and just default the `imageType` to `url`.

nylon22:
> Maybe we just remove the isRequired flag from SpacesImage and just default the imageType to url.

Good idea. I'll update

GoPro16 (changes requested):
Just need to add that prop onto the prop defs for the `SpacesImage` in the docs for people to know that its defaulting to `imageType` of `url`.
https://deploy-preview-321--availity-react-preview.netlify.com/components/spaces/images/

GoPro16 (approved):

PR #322, opened 2019-10-05
refactor(form): add unique id to radio and checkbox

### LINKED ISSUE #258: Form Radio And Checkbox | non string value blows up prop types
### Current Behavior

When a non boolean value is passed in as the value for a `checkbox` or `radio` it is also expected that we have an `id` to pass as the name of the `Input` and the `for` prop of the `FormGroup`.

When we do this 2 propType errors occur.
1 - FormGroup => `for` prop expected string
2 - Input => `name` prop expected string

### Expected behavior
You should be able to pass in any value that you want for checkbox/radio to accep and a custom `name` prop that you can just use for keeping the field names separated.

### Reproducible example

https://codesandbox.io/s/reactstrap-formik-issue-258-ucnkx

### Suggested solution(s)

1 - We update the Radio/Checkbox to pass `for` and `name` as `id` if it exists and make a note in the readme about `non-string` values

2 - We update the Radio/Checkbox to have a `name` prop instead of `id` and if it exists to overwrite as the `for` and `name` prop of the input/formgroup. BREAKING CHANGE

### Additional context

I think it made more sense keeping the props as the `name` because that is what everyone is used to but doing it with `id` was probably the wrong thing to do

stephanrotolante:
fixes #258

### DIFF
--- a/packages/form/src/Checkbox.js
+++ b/packages/form/src/Checkbox.js
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState } from 'react';
 import PropTypes from 'prop-types';
 import { Label, Input } from 'reactstrap';
 import uuid from 'uuid/v4';
@@ -15,7 +15,7 @@ const Checkbox = ({
 }) => {
   const { value, toggle, metadata } = useCheckboxGroup(checkValue);
 
-  const inputId = id || uuid();
+  const [inputId] = useState(id || uuid());
 
   const classes = classNames(
     className,
--- a/packages/form/src/Checkbox.js
+++ b/packages/form/src/Checkbox.js
@@ -1,6 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { Label, Input } from 'reactstrap';
+import uuid from 'uuid/v4';
 import classNames from 'classnames';
 import { useCheckboxGroup } from './CheckboxGroup';
 import FormGroup from './FormGroup';
@@ -14,6 +15,8 @@ const Checkbox = ({
 }) => {
   const { value, toggle, metadata } = useCheckboxGroup(checkValue);
 
+  const inputId = id || uuid();
+
   const classes = classNames(
     className,
     metadata.touched ? 'is-touched' : 'is-untouched',
@@ -21,10 +24,10 @@ const Checkbox = ({
   );
 
   return (
-    <FormGroup for={checkValue} check inline disabled={attributes.disabled}>
+    <FormGroup for={inputId} check inline disabled={attributes.disabled}>
       <Input
-        id={id || checkValue}
-        name={checkValue}
+        id={inputId}
+        name={inputId}
         className={classes}
         type="checkbox"
         {...attributes}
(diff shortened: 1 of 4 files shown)


nylon22 (changes requested):
Changes LGTM 🚀 Could we get a few simple tests, however? Should be very similar to the tests in this PR: https://github.com/Availity/availity-react/pull/308/files#diff-fdc295e0a32535d66153705e5597f8cc

robmcguinness:
Any chance we can switch to https://www.npmjs.com/package/nanoid. We use this lib internally within Node projects and also the bundle size is a bit smaller: https://bundlephobia.com/result?p=nanoid@2.1.2 vs https://bundlephobia.com/result?p=uuid@3.3.3

@GoPro16 @nylon22  thoughts?

GoPro16:
That should be fine. TBH its not like we need a super crazy uuid here. Just need something small enough to create a unique id for a field

stephanrotolante:
Why import a package that creates a uuid, when you can just make simple function within project that generates a string hash?

robmcguinness:
@stephanrotolante that is fine too. We have done this before https://github.com/Availity/availity-angular/blob/master/src/core/utils/uuid.js (copied from Angular source).

packages/form/src/Checkbox.js
```diff
@@ -14,25 +15,27 @@ const Checkbox = ({
 }) => {
   const { value, toggle, metadata } = useCheckboxGroup(checkValue);
 
+  const inputId = id || uuid();
```
TheSharpieOne: Wouldn't this potentially generate a new uuid with every render?

Maybe there is (or we can make) a `useUuid()` which will generate the uuid on the first render and return the same value every subsequent render. (or `useId()` to be agnostic to the underlying lib if you go with nanoid)

GoPro16: Actually, you are correct this most certainly will. Love the suggestion! We should just add this as a sub component under the `@availity/form` package as it could be useful for DX for people to use that so they can track the idea in a HOC for setting field values manually later.

I am trying to get away from having our own hooks package as their are a lot of other good mono repos out there that we could be using now for hooks. Would be a good feature request to have in something like [react-use](https://github.com/streamich/react-use)

TheSharpieOne: I suppose `const [id] = useState(() => uuid())` (or maybe even just `const [id] = useState(uuid)`, since `uuid` is a function and `useState` will not pass any args) would be similar. Doesn't seem like a specialized hook would be needed.

GoPro16: Right, it's a simple implementation but a pretty useful hook nonetheless that can just be an addition to some hooks package rather than defining it ourselves.

being way to overkill here with my comments so you guys can just ignore me 🙃

TheSharpieOne: I thought about adding it to react-use, but either have to pick the lib (uuid vs nanoid) or if we allow the developer to supply their own (to be more flexible), it ends up being a pointless abstraction (IMO) `const id = useId(uuid)` vs `const [id] = useState(uuid)`

nylon22 (approved):
LGTM. Thanks for contribution! 🙌

GoPro16 (approved):

PR #332, opened 2019-10-10
refactor(date): added test and pre-set date ranges

### LINKED ISSUE #331: Date Range does not have presets
### Current Behavior

**DateRange** in component `@availity/form` doesn't have the preset ranges

### Desired Behavior

When adding the ranges in the props, the component should be able to add the ranges button

### Suggested Solution

https://github.com/Availity/availity-react/blob/master/packages/reactstrap-validation-date/src/AvDateRange.js#L445

All users

Copy **AvDateRange** format

### Additional context

https://github.com/Availity/availity-react/blob/master/packages/reactstrap-validation-date/src/AvDateRange.js#L445

stephanrotolante:
fixes #331

### DIFF
--- a/packages/date/tests/DateRange.test.js
+++ b/packages/date/tests/DateRange.test.js
@@ -292,19 +292,32 @@ describe('DateRange', () => {
         }}
         onSubmit={onSubmit}
       >
-        <DateRange id="dateRange" name="dateRange" range />
+        <DateRange id="dateRange" name="dateRange" ranges />
         <Button type="submit">Submit</Button>
       </Form>
     );
 
     container.querySelector('.DateRangePickerInput_calendarIcon').click();
 
-    setTimeout(() => {
+    await wait(() => {
       expect(
         container.querySelector(
           '.DayPicker_calendarInfo__horizontal DayPicker_calendarInfo__horizontal_1'
         )
-      ).toBeTruthy();
-    }, 3000);
+      ).toBeDefined();
+    });
+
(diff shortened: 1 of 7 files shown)


packages/date/tests/DateRange.test.js
```diff
@@ -281,4 +281,30 @@ describe('DateRange', () => {
...
+        onSubmit={onSubmit}
+      >
+        <DateRange id="dateRange" name="dateRange" range />
+        <Button type="submit">Submit</Button>
+      </Form>
+    );
+
+    container.querySelector('.DateRangePickerInput_calendarIcon').click();
+
+    setTimeout(() => {
```
GoPro16: Instead of calling `setTimeout` you can use the methods from `@testing-library/react` to wait for the change

```js
import { wait } from '@testing-library/react'

await wait(() => {
expect(somethingToHappen)
})
```

GoPro16 (changes requested):
We still need to add the new prop typescript definition to the [docs here](https://availity.github.io/availity-react/form/date/components/date-range/#gatsby-focus-wrapper).

It looks like we didn't add any typescript definitions for the ranges prop as well in the `reactstrap-validation-date`. We probably should get that added in there as well.

Additionally would be good if we can grab one of the range items and simulate clicking it in a unit test to make sure the `datepicker` does get updated properly.

GoPro16 (approved):
LGTM

PR #335, opened 2019-10-11
feat(select,reactstrap-validation-select): Handle AutoFillType function values

### LINKED ISSUE #334: (select, reactstrap-validation-select) Allow function in values of AutoFillType
### Current Behavior

`autofill` can be a boolean or an object containing key/value pairs of a key corresponding to the `name` of an input on the form and a string value corresponding to a path on the selected option value to autofill like so:

```jsx
<Select
    ...
    autofill={{ payerName: 'payer.name', payerId: 'payer.id' }}
/>
```

### Desired Behavior

Handle the values of keys in an `AutoFillType` as a function for more complex logic on what should be autofilled. For example, say you have a `payer` input that should be autofilled to `payer.name` if it exists, otherwise it should be filled to `payer.id`. Would look like this:

```jsx
<Select
    ...
    autofill={{ payer: opt => opt.payer.name || opt.payer.id }}
/>
```

### Suggested Solution

For each `autofill` value, if it is a string, use `lodash.get` to grab the value. If it is a function, call it with the selected object and assign the returned result to the value to prefill

All

NA

### Additional context

NA

nylon22:
closes #334

### DIFF
--- a/docs/source/form/select/components/select.md
+++ b/docs/source/form/select/components/select.md
@@ -76,6 +76,16 @@ If `true`, when the value of the dropdown changes, if the `isMulti` prop is `fal
 
 For example, if the new value is `{ "payer": "Availity" }`, in order for the payer input in the form to be auto-filled to "Availity", the `name` prop on the input must be "payer". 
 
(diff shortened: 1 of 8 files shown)


GoPro16 (approved):
LGTM 🚀

PR #336, opened 2019-10-11
Refactor/babel

GoPro16:
- Removed `url` prop from `AvLink` as it was changed to use `href`.
- Fixed `submitted` state on `AvForm` causing stories to crash and refactored to function component
- Simplified babel preset by just using single `babel-preset-react-app`.


packages/link/Link.js
```diff
@@ -36,11 +36,10 @@ AvLink.defaultProps = {
 };
 
 AvLink.propTypes = {
-  url: PropTypes.string.isRequired,
   target: PropTypes.string,
   tag: PropTypes.oneOfType([PropTypes.func, PropTypes.string]),
   children: PropTypes.node,
-  href: PropTypes.string,
+  href: PropTypes.string.isRequired,
```
nylon22, line 42: This is a breaking change, no?

GoPro16: I just released the `href` change a few hours ago and needs to be set to required. Required prop types are just warnings and don't break the apps from what I can tell.

nylon22: People are currently using this component like:
```
<AvLink url="my-url-here"
```

I don't see anywhere in this change where `url` gets picked up from props and renamed to `href`.

jordan-a-young:
My understanding was that the name of the prop changed which would be a
breaking change.

packages/link/Link.js
GoPro16: https://github.com/Availity/availity-react/blob/master/packages/link/CHANGELOG.md

https://availity.github.io/availity-react/components/link/#href-string

nylon22: mea culpa. didn't see BREAKING CHANGE in commit message

nylon22 (approved):

PR #343, opened 2019-10-15
Refactor/favorites event

GoPro16:
- Fixed `onChange` callback to actually return the new state of the favorite change.
- Added `event: MouseEvent` to the callback in order to do things like `stopPropagation`
- Added Typescript Definitions


nylon22 (approved):
LGTM

PR #346, opened 2019-10-18
feat(analytics): added analytics package

GoPro16:
Adds wrapper around `@availity/analytics-core` package.

Makes it much easier to configure analytics instead of having to define it all using base javascript.

React also allows the component to unmount, removing all event listeners when developing locally. This prevents the number of API calls from being duplicated once the DOM Hot reloads after making changes.

I'm up for suggestions, I was thinking maybe for the React side we just allow a boolean or enum for what log message service you want to use and create it for the user rather than still having them provide the array themselves. We can still optionally allow them to pass in their own but feel like the same logic is duplicated across all our apps.

### DIFF
--- a/docs/gatsby-config.js
+++ b/docs/gatsby-config.js
@@ -66,6 +66,13 @@ module.exports = {
             },
           ],
           'Portal Components': [
+            {
+              resolve: 'components/analytics/index',
+              pages: [
+                'components/analytics/analytics',
+                'components/analytics/hook',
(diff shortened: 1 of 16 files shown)


jordan-a-young:
This looks awesome!

jordan-a-young (approved):

chrishavekost (approved):
LGTM, I like the `cleanup` effect too.

PR #353, opened 2019-10-21
feat(spaces): add useSpaces hook

nylon22:
Removes `useSpace` hook in favor of `useSpaces` hook, which returns a tuple of spaces. Unlike the `useSpace` hook, `useSpaces` does not return `loading` or `error` information. That information should be grabbed from `useSpacesContext`. 

BREAKING CHANGE: `useSpace` is no longer available. Use `useSpaces`
instead. Additionally, `isGhost` is no longer exported as that property is computed server side and assigned to the space asset

## Migrating

*Before:*
```js
import { useSpace } from '@availity/spaces';

// ...
const { space: space1, loading, error, isGhost } = useSpace('1');
const { space: space2 } = useSpace('2');
const { space: space3 } = useSpace('3');
```

*After:*
```js
import { useSpaces, useSpacesContext } from '@availity/spaces';

// ...
const [space1, space2, space3] = useSpaces('1', '2', '3');
const { loading, error } = useSpacesContext();
console.log('Space 1 is ghost', space1.isGhost);
```


chrishavekost (approved):

packages/spaces/index.d.ts
```diff
@@ -1,4 +1,4 @@
-import Spaces, { useSpace, SpacesContext } from './types/Spaces';
+import Spaces, { useSpaces, SpacesContext } from './types/Spaces';
```
GoPro16: We need to create a quick def for `useSpacesContext`

packages/spaces/src/Spaces.js
```diff
@@ -44,6 +44,8 @@ export const getAllSpaces = async (
 
 export const SpacesContext = createContext();
 
+export const useSpacesContext = () => useContext(SpacesContext) || {};
```
GoPro16: I'm not sure I agree with defaulting the method to return `{}`. If a user messed up implementing this, they would probably assume the context worked but didn't get any values back, rather than just returning `undefined` when the `SpacesProvider` doesn't exist.

packages/spaces/types/Spaces.d.ts
```diff
@@ -17,14 +17,15 @@ export interface SpacesProps {
 export interface SpaceType {
   error?: string;
   space: object;
-  isGhost: boolean;
   loading: boolean;
 }
 
-declare function useSpace(id?: string): SpaceType;
+declare function useSpaces(ids?: string[]): object[];
```
GoPro16: `useSpaces` declaration for the `...args` would be `...ids?: string`

I would create a `SpaceType` that at least contains the intellisense for `isGhost` so we know that is a valid value. In the future, I do see us having some global typescript definitions for our different domain models so probably a good place to start.

```ts
interface Space {
   [key:string]:any; // shortcut to allow any key/value
   isGhost?: boolean // adds intellisense for `isGhost`
}

GoPro16 (changes requested):
I think we are moving in the right direction with `@availity/spaces`. 🚀 📈

packages/spaces/index.d.ts
nylon22: I created it just forgot to export it 🤦‍♂. good catch: 316c69f

packages/spaces/src/Spaces.js
nylon22: agreed. refactored: ff148cb7aa18233230598fdf3869de0e0c486fda

packages/spaces/types/Spaces.d.ts
nylon22: this is what the ts linter recommended 🤷‍♂ c0e5db88321ca406ffe197e64414c0ae9020bdea

nylon22: went ahead and created `Space` type: 86cc811cac3f2b41f064aae3bc8fadab4d696ea6

GoPro16 (approved):
LGTM. Just made a one last update to tsdef for `SpacesContext`

PR #354, opened 2019-10-22
Fix/accessibility

chrishavekost:
This PR matches the disabled state colors of `Select.js` to those found in Availity/form. Also brings them into 508 compliance since there was not enough contrast with the previous values.

Before:
[Contrast Ratio 3.52:1](https://webaim.org/resources/contrastchecker/?fcolor=808080&bcolor=F2F2F2)
image

After:
[Contrast Ratio 6.89:1](https://webaim.org/resources/contrastchecker/?fcolor=495057&bcolor=E9ECEF)
image


stevenviar (commented):
LGTM

nylon22 (approved):
LGTM

PR #356, opened 2019-10-22
feat(date): add autoSync prop to allow for automatically filling dateranges

jordan-a-young:

8. Make sure your code passed the conventional commits check. Read more about [conventional commits](https://www.conventionalcommits.org/en/v1.0.0-beta.4/#summary)


TheSharpieOne:
🤔 `autosync` seems to describe what it does more than `autofill` based on how it handles when both values are present.
Also, the whole change the end date when the start date is _after_ the end date doesn't seem to intuitive.
You are forcing the user to change the start date first. Other implementations would simply swap the start and end dates if they were not chronological.
autofill-date-expectations

jordan-a-young:
@GoPro16 mentioned he was concerned about 508 compliance as well. I may need to rethink how this works. There seem to be more negatives than positives

packages/date/src/DateRange.js
```diff
@@ -28,7 +28,7 @@ const AvDateRange = ({
   datepickerProps,
   'data-testid': dataTestId,
   datepicker,
-  autofill,
+  autosync,
```
GoPro16: Should it be `autoSync` or `autosync`? `Autofill` made sense bc that is defined as a single word but not the other way? 📖 🤔 

Either is good with me but figured I would note in case you thought the same

nylon22: 🤓

nylon22: 8a6220b6bedee839a0435fdf955fd9cf091fddab

packages/date/src/DateRange.js
```diff
@@ -132,10 +133,31 @@ const DateRange = ({
...
+            [endKey]: value,
+          },
+          true
+        );
+        await setFieldTouched(name, true);
+      }
     }
+  };
+
+  const onFocusChange = async input => {
```
GoPro16, line 158: When `autoSync` is `true` and we auto fill then end date, we should also close the calendar if the user tabs off of the end date input potentially to prevent it from remaining open until the user has clicked out of the input.

GoPro16 (changes requested):

packages/date/src/DateRange.js
nylon22: I cannot find a way to hook into the `onBlur` event of the end date input.

The `<DateRangePickerInput />` does support an `onEndDateTab` prop. https://github.com/airbnb/react-dates/blob/7c58d6d6aadf0061980f1adeb52465c2606d58ac/src/components/DateRangePickerInput.jsx#L278

However, the `<DateRangePickerInputController />`, which is what the `<DateRangePicker />` renders, does not pass that prop down:

https://github.com/airbnb/react-dates/blob/7c58d6d6aadf0061980f1adeb52465c2606d58ac/src/components/DateRangePickerInputController.jsx#L311-L352

[Here](https://github.com/airbnb/react-dates/blob/master/src/shapes/DateRangePickerShape.js) are all the props that the `<DateRangePicker />` supports. 

I'll open up an issue on `react-dates` unless you see something I'm missing.

GoPro16: I think that should help us out here, more so than the `onFocusChange` potentially. Yeah it does look like its missing from the props. We can open up an issue and once that gets fixed we can re-open a new PR to fix this then as its a one off case with the focus on tab change.

GoPro16 (approved):

PR #358, opened 2019-10-23
refactor(form): fixed default checked when page loads if true

### LINKED ISSUE #357: @availity/form checkbox checked prop value should match field value
### Current Behavior
`<Input type="checkbox" />` when passed an `initialValue` of `true` should display the checkbox as checked. However, it doesn't.
### Expected behavior

A checkbox Input should allow the checkbox to be filled in with the `initialValue` is `true`

### Reproducible example

https://codesandbox.io/s/reactstrap-formik-checkbox-900pf

### Suggested solution(s)

Add logic to check the `type` in the `Input` component to add the `checked` value onto the reactstrap `Input`.

GoPro16:
fixes #357 

When `initialValues` for a input of type `checkbox` is `true` the checkbox checked prop should be set to `true` for it to display properly on the DOM.

### DIFF
--- a/packages/form/src/Input.js
+++ b/packages/form/src/Input.js
@@ -18,7 +18,15 @@ const Input = ({ tag: Tag, className, name, ...rest }) => {
       'was-validated'
   );
 
-  return <Tag className={classes} name={name} {...field} {...rest} />;
+  const extraProps = {};
+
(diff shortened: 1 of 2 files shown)


andrectoi:
Looks good to me!

andrectoi (approved):

PR #362, opened 2019-10-24
refactor(date): fixed error after picker selection

### LINKED ISSUE #360: DateField component validation not firing off correctly
## 🐛 Bug report

When using the `<DateField />` component with a yup schema, the validation appears to fire off incorrectly.

For example, I have the following yup schema:
```
validationSchema={yup.object().shape({
        dob: yup
          .date()
          .required("Please enter a valid date of birth.")
          .max("10/22/2019", "Please enter a date in the past.")
      })}
```
... which works in conjunction with this DateField component:
```
<DateField
        name="dob"
        calendarIcon={<FaCalendar />}
        label="DOB"
        max="10/22/2019"
/>
```

### Current Behavior:
If I choose a seemingly valid date such as 10/21/2019, the validation incorrectly marks it invalid. But then if I select the same date again (by either choosing it on the calendar widget or refocusing and unfocusing the input), it will correctly mark the field valid.

The same (but inverse) behavior happens when i select an invalid date such as 10/25/2019. It will incorrectly mark the field valid upon the first try, but then if I select the invalid date again, it will correctly mark the field invalid the second time.

### Expected Behavior:
The correct validation should run properly the first time the field loses focus, IMO. And if the field has already been touched, it should correctly validate on the fly.

### Reproducible example

Can see the above described behavior with this codesandbox link...
https://codesandbox.io/s/reactstrap-formik-scd4m?fontsize=14

### Suggested solution(s)

Unsure of a solution at this time. Still looking into the issue.

```bash
```

```
  Binaries:
    Node: 12.13.0 - ~/.nvm/versions/node/v12.13.0/bin/node
    Yarn: 1.19.1 - ~/.homebrew/bin/yarn
    npm: 6.12.0 - ~/.nvm/versions/node/v12.13.0/bin/npm
```

GoPro16:
For some reason, when we set the field value we also must set the field touched to true to essentially double validate. There maybe a bug in formik but this does resolve the issue and DateRange is performing the same logic without knowing that bug existed so it may be the correct way.

closes #360

### DIFF
--- a/packages/date/src/Date.js
+++ b/packages/date/src/Date.js
@@ -73,7 +73,9 @@ const AvDate = ({
       val = val.format(isoDateFormat);
     }
 
-    await setFieldValue(name, val, false);
+    await setFieldValue(name, val, true);
+
+    await setFieldTouched(name, true);
 
     if (onChange) {
       onChange(val);


nylon22 (approved):
LGTM 🙌

andrectoi (approved):
Looks good to me too!

PR #366, opened 2019-11-01
feat(spaces): add spaces link

GoPro16:
This is a draft PR as there are no complex components for link yet.

However, if you navigate to the storybook preview you will see that you can swap between the different use cases, 2 of which are legacy and will be removed eventually.

TODOS:
- [ ] 3 Types of Links ( Link what we do for SpacesImage )
- [ ] Unit test more. ( TopApp Updates need to be tested heavily )
- [x] Update Gatsby Docs
- [ ] Provide a better example in storybook

Additionally I am thinking we can bring in `@availity/favorites` as a `peerDep` and if we have already wrapped our app with it somewhere then one of the `Link` types can also render if the link is favorited or not next to it.

[SpacesLink Storybook](https://deploy-preview-366--availity-react-preview.netlify.com/storybook/?path=/story/components-spaces--spaceslink) ( You have to click the text for it to perform the link logic, its barebones at the moment )

### Canary Install
```
npm install @availity/spaces@canary @availity/page-header@canary
```


packages/spaces/src/Spaces.js
```diff
@@ -1,27 +1,56 @@
...
+  );
+};
 
 export const getAllSpaces = async (
   query,
   clientId,
   variables,
-  _spaces = []
+  _spaces = [],
+  mutliPayerRequired
```
nylon22: this is triggering my OCD 😆

packages/spaces/src/Spaces.js
```diff
@@ -91,25 +123,41 @@ const Spaces = ({
...
         const spacesBySpaceIDs = await getAllSpaces(
           query,
           clientId,
           vars,
-          spaces
+          spaces,
+          mutliPayerRequired
+        );
+
+        console.log("spacesBySpaceIDs",spacesBySpaceIDs);
```
nylon22: s/b removed before merge

packages/spaces/src/Spaces.js
```diff
@@ -175,41 +234,26 @@ Spaces.propTypes = {
...
+        ... SpaceFragment
+      }
+      payerSpaces: spaces(childIDs: $ids, payerIDs: $payerIDs, types: ["space"], page: $page, region: $payerSpaceRegion) @include(if:$includeParents){
+        ... SpaceFragment    
       }
     }
+    ${SpacesFragment}
   `,
-  variables: { types: ['space'] },
+  variables: { types: ['space'], includeParents: false },
```
nylon22, line 281: is there a corresponding change to `slotmachine` that supports this variable?

packages/spaces/src/SpacesLink.js
```diff
@@ -0,0 +1,135 @@
...
+  if(loading) return null;
+
+  const isNew = activeDate => dayjs().diff(activeDate,'day') < 30
+  return (
+    <Card title={name} className="mb-4 application" tabIndex={0} tag={CardLink} {...props} {...rest} aria-label={name}>
+      <CardBody className="card-block d-flex">
+        <span className="d-table-cell align-middle pr-2">
+          {/* <FavoriteHeart id={id} onChange={(_, e) => e.stopPropagation()} /> */}
+        </span>
+        <AppIcon className="d-table-cell align-middle mr-2">AI</AppIcon>
```
nylon22: s/b `shortName` from space i think instead of "AI"

packages/spaces/src/SpacesLink.js
```diff
@@ -0,0 +1,135 @@
...
+  const { loading } = useSpacesContext();
+  const [{ name, parents, metadata, description, activeDate } = {}, props = {}] = useLink(spaceId);
+  if(loading) return null;
+
+  const isNew = activeDate => dayjs().diff(activeDate,'day') < 30
+  return (
+    <Card title={name} className="mb-4 application" tabIndex={0} tag={CardLink} {...props} {...rest} aria-label={name}>
+      <CardBody className="card-block d-flex">
+        <span className="d-table-cell align-middle pr-2">
+          {/* <FavoriteHeart id={id} onChange={(_, e) => e.stopPropagation()} /> */}
```
nylon22: i think should be conditionally rendered if `props.allowFavorite` is true

packages/spaces/src/Spaces.js
GoPro16: No this is only to use the `@include` directive provided by default for `graphql`

packages/spaces/src/SpacesLink.js
GoPro16: Yeah that is the end goal. We will have to note though, that you will need to have your `FavoritesProvider` somewhere up the tree in order to use this.

packages/spaces/src/SpacesLink.js
GoPro16: Yeah this is in heavy draft state at the moment. I was messing around with the styling before going to replace the names. We probably should be checking the types on them as well to make sure the `shortname` is only used for navigations and the logos for everything else.

GoPro16:
Instead of providing 3 different subsets of components, this is using 1 super component that you can just pass booleans to for configuring.

Still need to write unit tests but this has been published under a major canary version for use in some of our projects.

packages/spaces/src/SpacesLink.js
```diff
@@ -0,0 +1,340 @@
...
+import AppIcon from '@availity/app-icon';
+import Icon from '@availity/icon';
+import { useSpacesContext } from './Spaces';
+import { isFunction } from './helpers';
+import useLink from './useLink';
+import IconTiles from './Tiles';
+import Loader, { skeletonPropType } from './Loader';
+
+const isNew = activeDate => dayjs().diff(activeDate, 'day') < 30;
+const getDisplayDate = date => dayjs(date).format('MM/DD/YYYY');
```
nylon22, line 28: Thoughts on allowing custom date format ?

packages/spaces/src/SpacesLink.js
```diff
@@ -0,0 +1,340 @@
...
+      link,
+      ...restLink
+    } = {},
+    props = {},
+  ] = useLink(propSpace || spaceId, {
+    clientId: propsClientId,
+    ssoAttributes,
+  });
+
+  const getIconTitle = useCallback(() => {
```
nylon22, line 106: looks like this logic is duplicated in `SpacesIcon`. can we consolidate?

nylon22 (changes requested):

packages/feedback/src/FeedbackForm.js
```diff
@@ -45,7 +45,7 @@ const FeedbackForm = ({
 
     await avLogMessagesApi.info({
       surveyId: `${name.replace(/\s/g, '_')}_Smile_Survey`,
-      smileLocation: `${name}`,
+      feedbackLocation: `${name}`,
```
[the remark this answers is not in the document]
GoPro16: This was originally what the names were supposed to be and I think when we added this to OSS land it was already not the proper one.

At least in the old angular apps this was the proper attribute. Maybe our teams just need to update their splunk reports to use the new world attribute names

docs/source/components/spaces/link.md
```diff
@@ -0,0 +1,85 @@
...
+Tag to overwrite the root component rendered.
+
+### `bodyTag?: React.ComponentType | string;`
+Tag to overwrite the body component that renders the title, description and date values. It defaults to `CardBody` or `div` depending on the value of the `linkStyle` prop.
+
+### `linkStyle?: string`
+When passed in, provides predefined styles for the component. Possible values are `card`, and `list`.
+
+### `icon?: boolean`
+When `true`, will render an `@availity/icon` next to the title if present on the space.
```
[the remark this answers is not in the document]
GoPro16: 🎁  This is going to get whack if people start passing in their own `queries` bc they may not get `icons.navigation` etc. back even though they want to display it. 

Going to update the comment to denote that it will be part of the object in the `Spaces`

gregmartDOTin:
Closed via #886

PR #368, opened 2019-11-01
Add waitUntilFocused prop to select libs

nylon22:
Adds `waitUntilFocused` prop to `select` and `reactstrap-validation-select`. Opening as Draft PR since it needs tests.

Also contains a fix for one of our date tests. Without this change, the tests would only work if the date was two digits instead of one. 👀


packages/reactstrap-validation-select/AvResourceSelect.js
```diff
@@ -164,6 +168,11 @@ class AvResourceSelect extends Component {
       .catch(() => ({ options: [], hasMore: false }));
   };
 
+  onFocusHandler = (...args) => {
+    this.setState({ dropdownHasBeenFocused: true });
+    if (this.props.onFocus) this.props.onFocus(...args);
```
GoPro16, line 168: Can we add the `onFocus` cb to `ResourceSelect` as well?

packages/reactstrap-validation-select/AvResourceSelect.js
```diff
@@ -11,11 +11,15 @@ class AvResourceSelect extends Component {
     super(props);
     this.state = {};
     this.state.previousOptions = [];
+    this.state.dropdownHasBeenFocused = false;
   }
 
   select = createRef();
 
   loadOptions = (...args) => {
+    if (this.props.waitUntilFocused && !this.state.dropdownHasBeenFocused) {
```
GoPro16: I think instead of adding this "`requireSatisfied`" logic ( bc that is essentially what this is ) by itself at the top you can put logic into that `boolean` `requiredSatisfied` just to clean up the flow of `loadOptions`

GoPro16 (changes requested):

packages/reactstrap-validation-select/AvResourceSelect.js
nylon22: should already be there. Line 42

packages/reactstrap-validation-select/AvResourceSelect.js
nylon22: good idea

packages/reactstrap-validation-select/AvResourceSelect.js
GoPro16: 🙈

GoPro16 (approved):

PR #376, opened 2019-11-18
refactor(reactstrap-validation-date): get default value from formctrl

GoPro16:
Fetches the default value from the `FormCtrl` model if it exists.

### DIFF
--- a/packages/reactstrap-validation-date/src/AvDate.js
+++ b/packages/reactstrap-validation-date/src/AvDate.js
@@ -34,7 +34,7 @@ class AvDate extends Component {
     return null;
   };
 
-  constructor(props) {
+  constructor(props, context) {
     super(props);
 
     this.state = {};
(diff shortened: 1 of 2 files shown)


chrishavekost:
LGTM

chrishavekost (approved):

packages/reactstrap-validation-date/src/AvDate.js
```diff
@@ -48,6 +48,12 @@ class AvDate extends Component {
     }
 
     this.state.focused = false;
+
+    const { getDefaultValue } = context.FormCtrl;
+
+    if (getDefaultValue(props.name)) {
```
nylon22, line 54: should this explicitly check for `undefined` ? What if someone specifies the default value for an input as `false` ?

GoPro16: I was more so just following what the `AvDateRange` did with the default value. I think since this is for a `Date` field the user should be entering a `string` value regardless so we don't have to make that check.

nylon22 (approved):
LGTM

PR #377, opened 2019-11-18
feat(pagination): add render prop

### LINKED ISSUE #372: Add render prop to Pagination.Content in @availity/pagination
### Current Behavior

Pagination items are rendered as such: https://github.com/Availity/availity-react/blob/db5b85ff944edc0b1581f878a08e131a9fd5e74e/packages/pagination/src/PaginationContent.js#L103-L114

### Desired Behavior / Suggested Solution

Allow optional `render` prop that provides the `items` and `Component` in the callback to give developer more control over how pagination items get rendered

nylon22:
closes #372


docs/source/components/pagination/content.md
```diff
@@ -42,3 +42,6 @@ If `true`, renders pagination content inside an infinite scroll component
 
 ### `infiniteScrollProps?: InfiniteScrollProps`
 Only used when `infiniteScroll` is true. See [react-infinite-scroll-component](https://github.com/ankeetmaini/react-infinite-scroll-component#props)
+
+### `render?: ({ ...UsePaginationReturnType, ...PaginationContentProps }) => ReactNode`
```
nylon22: If there's a better way to define this, please let me know. We could explicitly put all the properties of the `usePagination` return type and all of the prop types for `PaginationContent` but I was concerned about updates being made to one and not the other.

GoPro16: I think I get my words backwards when I say `render` prop. I was more so thinking instead of a render prop we directly let you make the children have the props like: https://github.com/Availity/availity-react/blob/master/packages/spaces/src/Spaces.js#L132-L134

For `PaginationContent` I would just worry about the props needed to render that piece. If the user wants to do anything outside of that they can just build their own custom component at that point which can do everything they are looking for.

nylon22: 56e566a375c61f399fde2bab398e1f9a543de031

packages/pagination/src/PaginationContent.js
```diff
@@ -90,6 +94,22 @@ const PaginationContent = ({
     );
   }
 
+  const items =
+    page &&
```
GoPro16: You're going to want to create a memoized version of this in case there are unnecessary re-renders happening.

packages/pagination/src/PaginationContent.js
```diff
@@ -41,41 +79,7 @@ const PaginationContent = ({
         hasMore={hasMore}
         dataLength={allPages.length}
       >
-        {allPages &&
```
GoPro16: Not sure if there is an easy way to just create one memoized function that returns the items array for the proper wrapper.

GoPro16 (changes requested):

packages/pagination/src/PaginationContent.js
nylon22: c97eca6842d986fc5021400eec82b82386496802

GoPro16 (approved):
LGTM. 🙌

PR #393, opened 2019-12-09
refactor(pagination): allow pagination to be controlled via props

GoPro16:
Allows `Pagination` to be controlled from props rather than uncontrolled through local state.


chrishavekost (approved):
LGTM

TheSharpieOne (approved):

PR #396, opened 2019-12-13
fix(reactstrap-validation-date): if max date parameter is a string wi…

dnoler:
When comparing a date value to a specified max, if the max value is a string with no time parameter, then we need to set the time component to EOD. Otherwise, max will be set to midnight, and it may appear that the selected value is outside the allowed range if it has a time component after midnight.

### DIFF
--- a/packages/date/src/utils.js
+++ b/packages/date/src/utils.js
@@ -29,9 +29,18 @@ export const isOutsideRange = (min, max, format = 'MM/DD/YYYY') => day => {
     }
 
     if (typeof max === 'string') {
-      isOutsideMax = day.isAfter(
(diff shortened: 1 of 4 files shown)


GoPro16:
For some reason, if you click the `calendarIcon` the maxDate is not disabled. But if you click inside the input field, the `maxDate` is disabled.

dnoler:
@GoPro16 That's due to the value passed in by the widget for each date. When you select the icon, the time value is set at midnight. When you select the input, or select a date in the widget, the time value is set at noon.

GoPro16 (approved):
We should add the same logic for the formik date and then we should be gewd.

GoPro16 (changes requested):
Meant to request changes

GoPro16 (approved):

PR #403, opened 2019-12-17
Add defaultToOnlyOption prop

nylon22:
## Description
Adds prop to `ResourceSelect` in `@availity/select` to default the select value to the first item in the results if only one item is returned from the `resource`. `defaultToOnlyOption` only defaults the select value under the following conditions:

- `defaultToOnlyOption` is true
- There is only one option
- `waitUntilFocused` is false - Otherwise we would be defaulting the value when the user is interacting with the dropdown or has previously interacted with the dropdown
- This is the first time calling the resource for the current state of `cacheUniq`. Again, otherwise we would be defaulting the value when the user is interacting with the dropdown or has previously interacted with the dropdown

## Example Use Case
It is a common use case at Availity that if a user is only tied to one organization, that the organization input should be defaulted to that single organization so that the user does not have to select from a list of one.

## Notes
- I added this prop to `<ResourceSelect />` instead of `<Select />` because the developer should be able to do this themselves with the `defaultValue` prop in `<Select />` since they control the options
- It is safe to not "prep" the value in `ResourceSelect` like how `Select` does before setting the field value because the value should always be the "raw" value in a `ResourceSelect`. See: https://github.com/Availity/availity-react/blob/7453f544a915c34b5ffa0c95bce21a23a71cc337/packages/select/src/Select.js#L77-L79
- I also removed `AvPatientSelect` from `select` and `reactstrap-validation-select`. The `resource` that backs it was never released. I do not consider this a breaking change because no one must've been using it since it wouldn't work anyways
- I did not add the prop to `reactstrap-validation-select` due to support cost of maintaining both packages and to encourage people to switch to get the new features


GoPro16 (approved):
LGTM

TheSharpieOne (approved):

chrishavekost (approved):
LGTM

PR #406, opened 2019-12-18
refactor(select): allow for parameter overrides

### LINKED ISSUE #404: ResourceSelect | Allow For Parameter Overrides
### Current Behavior

When using `AvResourceSelect`, there additional parameters can be passed in via the props. However the `limit`, `offset`, and `q` params are appended by default with static keys.

### Desired Behavior

The limit, offset, and `q` parameters should be allowed to get modified to fit the mold of whatever API is being called. Specific use cases are for any API that is extending the `AvApi` or `AvProxyApi` as different developers have different API schemas for fetching.

### Suggested Solution

Modify the `parameters` prop to allow for a `function` to be passed in. This function will take the 3 default params that we append by default as arguments and then returned object should be the parameters appended to the request.

Ex.
```
<AvResourceSelect
   name="select"
  parameters={({ q, limit,  offset }) => ({
    search: q,
    page: (offset/limit) + 1
  })}
/>
```

Anyone using their own custom `AvApi` or `AvProxyApi` from `@availity/api-axios`.

Alternatives to this solution are either overwriting the `postGet` method that is called on the `AvApi` itself or just passing a custom object to the `resource` prop that is no an `@availity/api-axios` resource.

hnicbaker:
Closes #404


packages/select/src/ResourceSelect.js
```diff
@@ -81,6 +81,18 @@ const ResourceSelect = ({
         },
       };
 
+      if (typeof data.variables === 'function') {
```
GoPro16: We should be checking if `rest.parameters` is a function so that we know if we need to spread the params onto the  variables or call the function to spread.

packages/select/src/ResourceSelect.js
```diff
@@ -92,6 +104,19 @@ const ResourceSelect = ({
         ...rest.parameters,
       };
     }
+
+    if (typeof params === 'function') {
```
GoPro16: Same here, we should be checking if `rest.parameters` is a function and not params

packages/select/src/ResourceSelect.js
```diff
@@ -101,6 +126,7 @@ const ResourceSelect = ({
     } else {
       page = 1;
     }
+
```
GoPro16, line 139: I would add that same `if/else` logic for spreading the parameters here so that we have already appended `limit` and `offset` to the params. If the developer has to change the way the params look with limit or offset they can then do it with those variables as arguments to the function.

Ex.

```
parameters={({ limit, offset}) => {{
   page: (offset/limit)+1
})}
```

packages/select/src/ResourceSelect.js
```diff
@@ -81,6 +81,18 @@ const ResourceSelect = ({
         },
```
GoPro16, line 80: We actually should remove the original `...rest.parameters` spread above so we can check first if we need to spread or call the function.

GoPro16 (changes requested):

GoPro16 (commented):
Let's add 1 more test for the graphql based select so we can get the test coverage back up.

GoPro16 (changes requested):
Last thing we need to do is add this same logic in for `@availity/reactstrap-validation-select` pkg.

GoPro16 (approved):
Nice! LGTM

PR #408, opened 2019-12-20
add form-upload package

nylon22:
This is the equivalent of `@availity/upload`, which is meant for use with `availity-reactstrap-validation`, but for use with `@availity/form`.

The only major difference is some of the state management in `@availity/form-upload` is more tightly coupled to `formik`.


packages/feedback/package.json
```diff
@@ -42,7 +42,7 @@
     "@availity/api-core": ">=2.0.0",
     "@availity/localstorage-core": "^2.6.1",
     "axios": "^0.18.0",
-    "formik": "^2.0.1-rc.5",
+    "formik": "^2.0.0",
```
TheSharpieOne: this was downgraded, any reason was the RC was needed before that might affect this?

packages/form-upload/FileRow.js
```diff
@@ -0,0 +1,99 @@
...
+  const ext = file.file.name
+    .split('.')
+    .pop()
+    .toUpperCase();
+
+  const icon = fileTypeIconMap[ext.toLowerCase()] || 'doc';
+
+  if (typeof children === 'function') {
+    return children({
+      file,
```
GoPro16, line 43: Instead of having `children` as a function for the typescript definition can we add these all as named arguments to the function that returns void in the definition. This way people can intellisense and also not have errors when rendering children with a function.

packages/feedback/package.json
nylon22: Good catch. `2.0.0` is deprecated, anyways. i'll change to `^2.0.1`

GoPro16 (approved):

PR #409, opened 2019-12-20
Add modalHeaderProps to FeedbackForm

nylon22:
Adds new prop called `modalHeaderProps` to `<FeedbackForm />`, which is spread onto the `<ModalHeader/>` rendered inside of the `<FeedbackForm />`. 

This is especially useful for overriding the default `tag` that is rendered by `<ModalHeader />`, which [defaults to h5](https://github.com/reactstrap/reactstrap/blob/master/src/ModalHeader.js#L19). Depending on the app, this prop may be needed for accessibility: https://www.w3.org/WAI/tutorials/page-structure/headings/

### DIFF
--- a/packages/feedback/typings/FeedbackForm.d.ts
+++ b/packages/feedback/typings/FeedbackForm.d.ts
@@ -8,15 +8,16 @@ type AboutOption = {
     label?: React.ReactNode;
 };
 
-export interface FeedBackFormProps {
+export interface FeedbackFormProps {
     name: string;
     onFeedbackSent?: (feedback?: {[key:string]:any}) => void;
     faceOptions?: FaceOption[];
     aboutOptions?: AboutOption[];
     prompt?: string;
     staticFields?: object;
+    modalHeaderProps?: object;
 }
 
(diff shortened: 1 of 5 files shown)


packages/feedback/typings/FeedbackForm.d.ts
```diff
@@ -8,15 +8,16 @@ type AboutOption = {
...
 
-export interface FeedBackFormProps {
+export interface FeedbackFormProps {
     name: string;
     onFeedbackSent?: (feedback?: {[key:string]:any}) => void;
     faceOptions?: FaceOption[];
     aboutOptions?: AboutOption[];
     prompt?: string;
     staticFields?: object;
+    modalHeaderProps?: object;
```
GoPro16: Let's add `@types/reactstrap` to the dev deps of this package and then use the props to extend here.

ex.
```js
yarn add @types/reactstrap
...
import { ModalHeaderProps } from 'reactstrap'
...
modalHeaderProps?: ModalHeaderProps;
```

GoPro16 (approved):
LGTM

chrishavekost (approved):

PR #410, opened 2019-12-20
feat(link): default rel prop when linking to external sites

chrishavekost:
Provides `rel` with default value of `noopener noreferrer` when linking to external sites. Based on best practice detailed [here](https://developers.google.com/web/tools/lighthouse/audits/noopener).

This will only apply if `target = "_blank"`, `rel` has no specified value, the URL is `absolute`, `loadApp = false`, and the destination hostname is not the same as our current hostname.

`AvLink` has also been modified so that `getUrl` is only called once, and the `absolute` property of the `href` has been extracted for reuse.

### DIFF
--- a/packages/link/Link.js
+++ b/packages/link/Link.js
@@ -3,32 +3,54 @@ import PropTypes from 'prop-types';
 import { isAbsoluteUrl } from '@availity/resolve-url';
 
 // if absolute or loadApp is disabled, return url. otherwise loadappify the url
-export const getUrl = (url = '', loadApp) => {
-  const absolute = isAbsoluteUrl(url);
+export const getUrl = (url = '', loadApp, absolute) => {
   if (absolute || !loadApp) return url;
 
   return `/public/apps/home/#!/loadApp?appUrl=${encodeURIComponent(url)}`;
 };
 
+// takes href and transforms it so that we can compare hostnames and other properties
+const getLocation = href => {
+  const location = document.createElement('a');
+  location.href = href;
+  return location;
+};
+
 const AvLink = ({
   tag: Tag,
-  href: url,
+  href,
   target,
   children,
   onClick,
   loadApp,
+  rel,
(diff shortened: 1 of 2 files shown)


packages/link/Link.js
```diff
@@ -3,32 +3,54 @@ import PropTypes from 'prop-types';
...
-  >
-    {children}
-  </Tag>
-);
+}) => {
+  const absolute = isAbsoluteUrl(href);
+  const url = getUrl(href, loadApp, absolute);
+
+  // eslint-disable-next-line no-cond-assign
+  if ((target = '_blank') && !rel && absolute) {
```
GoPro16: Just wanted to make sure this is correct here. So the only time the first condition would fail is if `target === null`

GoPro16 (commented):
Tests look good, just want to confirm that is the only edge case we need to account for above.

packages/link/Link.js
nylon22: https://eslint.org/docs/rules/no-cond-assign has this to say:

> There are valid reasons to use assignment operators in conditional statements. However, it can be difficult to tell whether a specific assignment was intentional.

I think we should be explicitly checking if `target` is equal to `"_blank"`
```
if (target === '_blank' && ...

File: packages/link/Link.js [hunk 2]
```diff
@@ -3,32 +3,54 @@ import PropTypes from 'prop-types';
...
+
 const AvLink = ({
   tag: Tag,
-  href: url,
+  href,
   target,
   children,
   onClick,
   loadApp,
+  rel,
```
nylon22: rather than destructuring `rel` from props, leave it out, and if it's specified it will be spread onto the `Tag` and override whatever we set it to.

nylon22 (changes requested):

nylon22 (approved):
LGTM

PR #419, opened 2019-12-26
fix(feedback): Use formProps passed to Feedback component in underlying FeedbackForm

ckallemeyn:

### DIFF
--- a/packages/feedback/src/FeedbackForm.js
+++ b/packages/feedback/src/FeedbackForm.js
@@ -115,10 +115,22 @@ const FeedbackForm = ({
         validationSchema={yup.object().shape({
           feedback: yup
             .string()
+            .trim()
+            .matches(
+              /^([^\W])[a-zA-Z0-9!?.'"\s]+$/,
+              'Additional Feedback must contain alphanumeric characters'
+            )
+            .min(5, 'Additional Feedback must exceed 5 characters.')
             .max(200, 'Additional Feedback cannot exceed 200 characters.')
             .required('This field is required.'),
           additionalFeedback: yup
             .string()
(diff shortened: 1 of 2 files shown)


nylon22 (changes requested):
This is a breaking change as it contains changes some developers may not want. I think a better approach would be to allow the developer to provide a custom ‘validationSchema’ to the form as a prop.

packages/feedback/src/FeedbackForm.js
```diff
@@ -112,26 +113,29 @@ const FeedbackForm = ({
...
-              icon: yup.string().required(),
-              description: yup.string(),
-              placeholder: yup.string(),
-            })
-            .required('This field is required.'),
-          feedbackApp: yup
-            .string()
-            .isRequired(aboutOptions.length > 0, 'This field is required.'),
-        })}
+        validationSchema={
```
nylon22: On second thought, let’s add a new prop, ‘formProps’ that gets spread onto the Form, so that the developer can override any of the props for the Form, not just the ‘validationSchema’

packages/feedback/src/FeedbackForm.js
```diff
@@ -219,6 +221,7 @@ FeedbackForm.propTypes = {
   prompt: PropTypes.string,
   additionalComments: PropTypes.bool,
   staticFields: PropTypes.object,
+  validationSchema: PropTypes.object,
```
nylon22: Let's remove this since it's not explicitly used in the component

nylon22 (changes requested):

nylon22 (approved):
LGTM

PR #425, opened 2020-01-08
fix(select): field props interface made generic

djejaquino:
`SelectFieldProps` extends `SelectProps` without declaring its generic parameter. This results in a series of errors when declaring a `<SelectField>` on a Typescript project.

The right thing to do here is to make `SelectFieldProps` also generic, letting users who extend it define which type they want for their `selectRef?: React.Ref<T>`.


GoPro16 (approved):
LGTM. Appreciate the PR! 🙌

PR #431, opened 2020-01-22
docs: remove unneccessary semicolons following JSX tags

shkn-not-strrd:
docs: remove unnecessary semicolons following JSX tags in code examples, for better user experience

I was using the (very handy) copy-to-clipboard function in the documentation, and noticed that there were extra semicolons in several code examples that required cleaning up when I pasted into my project. I figured, why not save everyone the extra step?


GoPro16 (approved):
LOL. Appreciate your PR! Hopefully they don't get added back after auto formatter kicks in on other people's linters 😬

PR #435, opened 2020-01-31
fix(select): fixes case where validation messages do not appear due t…

### LINKED ISSUE #434: Passing an inputId to SelectField can break Feedback
### Current Behavior

`SelectField` currently checks if an `inputId` prop exists. If it does, and if `inputId !== name`,  then the user will not see form validation messages unless they define their `validationSchema` differently.

### Expected behavior

React-select sets the `inputId` automatically, so we do not need to check for it or associate it with our `Label`. Users should expect their `validationSchema` to use the `name` of the `SelectField`, and not have to change that if they pass a custom `inputId`.

### Additional context

There is also a small issue where passing an `id` prop to `SelectField` can cause an orphaned label and break 508 compliance when it is equal to `name`. React-select creates and defines an `id` prop internally. Currently, by spreading `attributes` over `Select`, the `id` gets assigned to `innerProps` of `SelectContainer`, which assigns the same `id` to `EmotionCssPropInternal`. As far as I can tell, this duplicate id conflicts with 508 compliance checkers and screen readers, causing orphaned label warnings to appear.

chrishavekost:
…o inputId being different than name. Closes #434

Warns about 508 compliance issue if a user passes an id === name

### DIFF
--- a/packages/select/src/SelectField.js
+++ b/packages/select/src/SelectField.js
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useEffect } from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
 import { FormGroup, Feedback } from '@availity/form';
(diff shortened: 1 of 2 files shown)


packages/select/src/SelectField.js
```diff
@@ -1,29 +1,31 @@
...
+    }
+  }, [attributes.id, name]);
+
   let thisLabel = false;
   if (label) {
-    let forLabel = name;
-    if (attributes.inputId) {
-      forLabel = attributes.inputId;
-    } else {
-      attributes.inputId = name;
```
GoPro16, line 23: Can we still provide the default setter for `inputId` when its value is undefined or will that break other flows?

Meaning
```js
attributes.inputId = attributes.inputId || name
```

chrishavekost: [React-select will set this by default](https://github.com/JedWatson/react-select/blob/master/packages/react-select/src/Select.js#L1404) so I think we are safe to leave this out of our code, unless we start using `inputId` for other purposes.

GoPro16: `inputId` is good for use during unit tests to grab the input by an id that you know of.

It looks like the `inputId` in their case is just [incremented by a count](https://github.com/JedWatson/react-select/blob/master/packages/react-select/src/Select.js#L367).

So passing in a `name` prop to or component will guarantee the `inputId` is:
1 - Unique
2 - The inputId in `react-select` is always `${name}-input`

GoPro16 (approved):

packages/select/src/SelectField.js
chrishavekost: Ah so actually it looks like we default `inputId` to `name` in [our Select component](https://github.com/Availity/availity-react/blob/master/packages/select/src/Select.js#L106-L108), so we should still be good.

PR #436, opened 2020-01-31
feat(select): add shouldSearch prop to ResourceSelect, defaultToCurrentRegion prop to AvRegionSelect

nylon22:
## Changes
- Adds `shouldSearch` prop to `ResourceSelect`, which can be a boolean or a function. It's similar to the `isDisabled` prop from `react-select`, except the `shouldSearch` prop gets passed the arguments to the `loadOptions` function. This is an important difference because it solves the use case of if the `loadOptions` should not be fired off if the `inputValue` doesn't match a certain pattern, for example.
- Adds `defaultToCurrentRegion` prop to `AvRegionSelect`, useful for if the value of the region input should be defaulted to the user's currently selected region


GoPro16 (changes requested):
Recalling our conversation about the hooks package I think its okay in this instance to use the hooks package since ware exporting this select component inside of `resources.js` and not the main `index.js` that the base component is exported from.

This block looks good but feel free to disregard what I said previously about the hooks pkg if you want to reduce the code footprint here.

GoPro16 (approved):
You will be a typescript developer in no time 😃

PR #442, opened 2020-02-07
Feat/phone

chrishavekost:
This PR adds a standard phone component for teams to use in their forms. The reason for it being placed in its own package is so that the bundle size of `@availity/yup` and `@availity/form` doesn't grow unnecessarily larger due to `libphonenumber-js`.

`libphonenumber-js` was chosen because it is heavily based on Google's phone number library, but comes in a smaller package.

The `Phone` component works with `yup.phone()` validation OOTB, and a new `validatePhone()` `yup` method has been written. It uses a relaxed validation by default ([benefits](https://github.com/catamphetamine/libphonenumber-js/blob/master/README.md#using-phone-number-validation-feature)), and strict validation can be used with `yup.string().validatePhone(errorMessage, true)`.

To aid in 508 compliance and a [better user experience](https://www.filamentgroup.com/lab/politespace.html), phone numbers (and partial numbers) are only formatted `onBlur`.

Along the way, while trying to resolve issues with `Jest` and unit tests, many packages across `availity-react` were updated. I believe I resolved the few breaking changes introduced.


packages/phone/typings/Phone.d.ts
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+import { Field } from '@availity/form/typings';
+
+interface ExtensionProps extends Field {
+  name: string;
+  id?: string;
+  extColProps?: object;
+}
+export interface PhoneProps extends Field {
```
GoPro16: `PhoneProps` should be extending [FieldProps](https://github.com/Availity/availity-react/blob/master/packages/form/typings/Field.d.ts#L6) not field.

storybook/stories/phone.stories.js
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import { withKnobs } from '@storybook/addon-knobs/react';
+
+import Phone from '@availity/phone';
+import README from '@availity/phone/README.md';
+
+storiesOf('TODO|Phone', module)
```
GoPro16: You should still put this under `Form` in storybook.

docs/source/form/phone/index.mdx
```diff
@@ -0,0 +1,16 @@
+---
+title: Phone
```
GoPro16, line 2: This still needs to be added to `gatsby-config.js`?

GoPro16 (changes requested):
A little scary that this package is quite large just for phone validation but should be okay if we are not using everything provided. Would be interesting to see a bundle size on our apps when its included and minified.

chrishavekost:
> A little scary that this package is quite large just for phone validation but should be okay if we are not using everything provided. Would be interesting to see a bundle size on our apps when its included and minified.

I agree, that's one reason I think it should be a separate package. You just reminded me though, I want to include an example in the doc of how to lazy load this with dynamic imports and `Suspense`, so that the larger size has less of an impact.

GoPro16 (approved):
LGTM

PR #450, opened 2020-02-12
fix: updates @availity/yup to fix resolution issues. Fixes issue with…

chrishavekost:
… Phone documentation not showing


packages/date/package.json
```diff
@@ -32,13 +32,15 @@
   },
   "peerDependencies": {
     "@availity/form": "^0.1.0",
+    "@availity/yup": "^2.0.0",
```
GoPro16, line 35: This is a breaking change and will need to be noted to major bump `@availity/date`

GoPro16 (changes requested):

GoPro16 (approved):
🙏

PR #457, opened 2020-02-18
feat(authorize): adds currentRegion to values returned from useAuthor…

chrishavekost:
…ize()

Useful for avoiding unnecessary calls to regions API if content is rendered under an `Authorize` component and depends on current region.

### DIFF
--- a/docs/source/components/authorize.mdx
+++ b/docs/source/components/authorize.mdx
@@ -19,12 +19,12 @@ import Authorize from '@availity/authorize';
 
 <Authorize permissions="1234">
   {/* stuff to render if the user is authorized with permission indicated above */}
-</Authorize>
+</Authorize>;
 ```
 
 ## Props
 
-### `permissions: string | number | Array<string | number | Array<string | number>>` 
+### `permissions: string | number | Array<string | number | Array<string | number>>`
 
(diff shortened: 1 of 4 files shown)


docs/source/components/authorize.mdx
```diff
@@ -19,12 +19,12 @@ import Authorize from '@availity/authorize';
 
 <Authorize permissions="1234">
   {/* stuff to render if the user is authorized with permission indicated above */}
-</Authorize>
+</Authorize>;
```
nylon22: @shkn-not-strrd did a lot of work to get rid of these. don't make her have to do it again 😆

packages/authorize/useAuthorize.js
```diff
@@ -105,6 +104,7 @@ export default (
     ) {
       setLoading(false);
       setAuthorized(authorized);
+      setCurrentRegion(currentRegion);
```
nylon22: I don't think we should set the current region conditionally. if it resolves, we should set it on line 84.

packages/authorize/useAuthorize.js
```diff
@@ -116,5 +116,5 @@ export default (
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [organizationId, region, customerId, permissions]);
 
-  return [authorized, loading];
+  return [authorized, loading, currentRegion];
```
nylon22, line 124: need to update TS def: https://github.com/Availity/availity-react/blob/master/packages/authorize/useAuthorize.d.ts#L13

packages/authorize/useAuthorize.js
```diff
@@ -7,6 +7,7 @@ export default (
 ) => {
   const [authorized, setAuthorized] = useState(false);
   const [loading, setLoading] = useState(true);
+  const [currentRegion, setCurrentRegion] = useState('');
```
nylon22, line 10: let's add a test that you can pull back the current region from the `useAuthorize` hook: https://github.com/Availity/availity-react/blob/master/packages/authorize/tests/useAuthorize.test.js#L55

nylon22 (changes requested):

nylon22 (approved):
LGTM

PR #458, opened 2020-02-18
bug(form): enable the usage of formik children props

### LINKED ISSUE #428: A way to access Formik children/component props
## ❓Question

Is there a way to access Formik [methods/props](https://jaredpalmer.com/formik/docs/api/formik#formik-render-methods-and-props) through `@availity/form`?

Two things to consider:

- The [children](https://github.com/Availity/availity-react/blob/1aaf6f8d87b497f73a1250f0a7c749034170e224/packages/form/src/Form.js#L29) of Availity `Form` are turned into `children` of Reacstrap `Form`; passing a function to it breaks the app and props are not injected (duh);
- Adding a `component` prop to Availity `Form` and passing it to `Formik` makes Reactstrap `Form` obsolete.

djejaquino:
Fixes #428


packages/form/src/Form.js
```diff
@@ -41,7 +41,7 @@ Form.propTypes = {
   initialTouched: PropTypes.object,
   validationSchema: PropTypes.object,
   validate: PropTypes.func,
-  children: PropTypes.node,
+  children: PropTypes.oneOf([PropTypes.func, PropTypes.node]),
```
GoPro16, line 44: Can you also update the `typescript` definition for this to work?

Example in `Spaces`:  https://github.com/Availity/availity-react/blob/master/packages/spaces/types/Spaces.d.ts#L82

GoPro16 (changes requested):

packages/form/src/Form.js
djejaquino: I think that is already done. `Form` [extends](https://github.com/Availity/availity-react/blob/db870082e5eb487c455963578390be6f6dfec816/packages/form/typings/Form.d.ts#L4) `React.Component` with `FormikConfig`, which [expects children as Function](https://github.com/jaredpalmer/formik/blob/f117c04738ed218b5eb8916d7189e0849962d50d/packages/formik/src/types.tsx#L178).

GoPro16: Oh true 🤦‍♂ 

I forgot to notice we already extend the `FormikConfig`

GoPro16 (approved):
LGTM. 🚀

PR #477, opened 2020-03-03
Fix upload drop rejection error messages not displaying

nylon22:
This PR introduces a breaking change to `form-upload` and `upload`. the `react-dropzone` peer dependency must now be `>= 11.x` in order for this package to work.

Fixes issue in `upload` and `form-upload` where error messages for rejected files are not displayed when `showFileDrop` is true

## Explanation of issue
1. Error is set here: https://github.com/Availity/availity-react/blob/master/packages/upload/Upload.js#L87
2. Then `setFiles` is called: https://github.com/Availity/availity-react/blob/master/packages/upload/Upload.js#L90
3. `setFiles` immediately clears out the error https://github.com/Availity/availity-react/blob/master/packages/upload/Upload.js#L76

The same flow occurs in `form-upload`

## Solution
This PR resolves this by attaching the rejection reason from `react-dropzone` to each file, and allowing `<UploadProgressBar />` to render the error message: https://github.com/Availity/availity-react/blob/master/packages/upload/UploadProgressBar.js#L66 for consistent error styling regardless of whether `showFileDrop` is used. (In `form-upload`, I left the `<Feedback />` so that if someone use the `useFormikContext` hook to explicitly set a the field's error it will still display.)

## Additional Info
- this PR passes the `allowedFileTypes` prop as the `accept` prop to `react-dropzone`
- this PR adds a `getDropRejectionMessage` prop so that user land can override error messages

### DIFF
--- a/docs/source/form/upload/upload.md
+++ b/docs/source/form/upload/upload.md
@@ -70,7 +70,7 @@ Indicates that the user will be allowed to select multiple files when selecting 
 
 Set as true to show a drag and drop file upload option instead of a button (file explorer still available on click).
 
-### `getDropRejectionMessage?: (errorCode: String, file: File) => String`
+### `getDropRejectionMessage?: (errors: Array<FileError>, file: File) => String`
 
 Override the default error message for files rejected when `showFileDrop` is `true`.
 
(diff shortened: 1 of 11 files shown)


GoPro16 (approved):
LGTM, not sure if you just want to add the change that was attempted in the below PR for `disabled` as well since its rather simple/short

https://github.com/Availity/availity-react/pull/523

nylon22:
> LGTM, not sure if you just want to add the change that was attempted in the below PR for `disabled` as well since its rather simple/short
> 
> #523

sounds good 👍 . i'll just go ahead and update that PR.

PR #484, opened 2020-03-13
Feat/date year picker

### LINKED ISSUE #469: Ability to change year faster in date picker
### Current Behavior

Date picker allows us to pick a date that is close to the current date but if we want to re use the date picker for date of birth, and the user's year of birth is for example 1965. The user would have to click the back arrow hundreds of times to reach to the desired year. 

### Desired Behavior

Date picker would allow for an optional way to "select the year" if they want to go back many years from the current date. 

Make the year a dropdown for easy access to any year. 

Not use the date picker at all, as it's not flexible.

chrishavekost:
Closes #469. 

Adds the ability to select both month and year values from a single date picker and a date range picker, allowing for quicker navigation through the picker when not inputting date via text.

Date - general

Adds ability for user to pass in a `customArrowIcon`, new default is the svg arrow from `react-dates`.

Moves `navPosition` to `navPositionBottom` so that we don't have to resort to a bunch of CSS styling to get a clean look.

`navPositionTop`:
Date - navPositionTop

`navPositionBottom`:
Date - navPositionBottom

Fixes styling issue in Firefox where component size was not uniform:

Before:
FF CSS issue

After:
FF CSS issue fixed

### DIFF
--- a/packages/date/src/Date.js
+++ b/packages/date/src/Date.js
@@ -27,7 +27,6 @@ const AvDate = ({
   format,
   validate,
   datePickerProps,
-  yearPickerProps,
   'data-testid': dataTestId,
   ...attributes
 }) => {
@@ -111,9 +110,7 @@ const AvDate = ({
   };
 
   const renderMonthElement = ({ month, onMonthSelect, onYearSelect }) => {
-    const { minYear, maxYear } = yearPickerProps;
-    // TODO: optimize call to buildYearPickerOptions
-    const yearPickerOptions = buildYearPickerOptions(minYear, maxYear, month);
+    const yearPickerOptions = buildYearPickerOptions(min, max, month, format);
     return (
       <Row>
         <Col>
@@ -207,17 +204,12 @@ AvDate.propTypes = {
   datepicker: PropTypes.bool,
   validate: PropTypes.func,
   datePickerProps: PropTypes.object,
-  yearPickerProps: PropTypes.object,
 };
 
 AvDate.defaultProps = {
   calendarIcon: <Icon name="calendar" />,
   format: 'MM/DD/YYYY',
   datepicker: true,
-  yearPickerProps: {
-    minYear: moment().year() - 100,
-    maxYear: moment().year(),
-  },
 };
 
 export default AvDate;
--- a/packages/date/src/Date.js
+++ b/packages/date/src/Date.js
@@ -116,13 +116,14 @@ const AvDate = ({
         <Col>
           <select
             data-testid="monthPicker"
+            aria-label="month picker"
             value={month.month()}
             onChange={e => {
               onMonthSelect(month, e.target.value);
             }}
           >
             {moment.months().map((label, value) => (
-              <option key={label} value={value}>
+              <option key={label} value={value} aria-label={label}>
                 {label}
               </option>
             ))}
@@ -131,6 +132,7 @@ const AvDate = ({
         <Col>
           <select
             data-testid="yearPicker"
+            aria-label="year picker"
             style={{
               minWidth: '100%', // 4 digit years not wide enough to fill column
             }}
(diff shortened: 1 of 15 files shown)


packages/date/src/Date.js
```diff
@@ -121,6 +171,7 @@ const AvDate = ({
         data-testid={`date-input-group-${name}`}
       >
         <SingleDatePicker
+          renderMonthElement={renderMonthElement}
```
chrishavekost, line 173: `renderMonthElement` is placed above `{...datePickerProps}` just in case people are already passing in `datePickerProps.renderMonthElement` on their own.

packages/date/src/DateRange.js
```diff
@@ -252,6 +308,7 @@ const DateRange = ({
         }}
       >
         <DateRangePicker
+          renderMonthElement={renderMonthElement}
```
chrishavekost, line 310: `renderMonthElement` is placed above `{...datePickerProps}` just in case people are already passing in `datePickerProps.renderMonthElement` on their own.

packages/date/src/DateRange.js
```diff
@@ -238,6 +245,55 @@ const DateRange = ({
     ) : null;
   };
 
+  const renderMonthElement = ({ month, onMonthSelect, onYearSelect }) => {
+    const { minYear, maxYear } = yearPickerProps;
+    // TODO: optimize call to buildYearPickerOptions
+    const yearPickerOptions = buildYearPickerOptions(minYear, maxYear, month);
```
chrishavekost: Performance seems fine with this, but can be optimized if needed.

packages/date/src/Date.js
```diff
@@ -109,6 +110,55 @@ const AvDate = ({
     return null;
   };
 
+  const renderMonthElement = ({ month, onMonthSelect, onYearSelect }) => {
+    const { minYear, maxYear } = yearPickerProps;
+    // TODO: optimize call to buildYearPickerOptions
+    const yearPickerOptions = buildYearPickerOptions(minYear, maxYear, month);
```
chrishavekost: Performance seems fine with this, but can be optimized if needed.

packages/date/src/Date.js
```diff
@@ -109,6 +110,55 @@ const AvDate = ({
     return null;
   };
 
+  const renderMonthElement = ({ month, onMonthSelect, onYearSelect }) => {
+    const { minYear, maxYear } = yearPickerProps;
```
GoPro16: Same thing here. Can we not just re-use the `min` and `max` props and derive the current year from there?

I would say if the minDate is Jan 1 then its safe to use that year and the max would be if December 31 is there then the max year would be in there or something to that nature.

GoPro16 (approved):
LGTM. Would be good if someone can also give it a second look.

packages/date/src/Date.js
```diff
@@ -109,6 +109,53 @@ const AvDate = ({
     return null;
   };
 
+  const renderMonthElement = ({ month, onMonthSelect, onYearSelect }) => {
+    const yearPickerOptions = buildYearPickerOptions(min, max, month, format);
+    return (
+      <Row>
+        <Col>
+          <select
+            data-testid="monthPicker"
```
nylon22, line 118: i believe these need `aria-label` attributes for screen readers

packages/date/src/DateRange.js
```diff
@@ -265,7 +319,7 @@ const DateRange = ({
           disabled={attributes.disabled}
           onFocusChange={onFocusChange}
           renderCalendarInfo={renderDateRanges}
-          customArrowIcon="-"
+          customArrowIcon={customArrowIcon}
```
nylon22, line 324: should this be added to default props to remain backwards compatible?

chrishavekost: I don't think it has to be because when `customArrowIcon` is `undefined` then react-dates provides the default svg arrow icon for us.

nylon22 (approved):

PR #488, opened 2020-03-17
Fix for Date Range autoSync infinite loop bug

### LINKED ISSUE #483: autoSync in DateRangeField causes browser window to lock up
Steps to reproduce:

1. Navigate to docs: https://availity.github.io/availity-react/form/date/components/date-range-field/
2. Add autoSync prop to the DateRangeField
3. Select both dates in the calendar picker - the calendar will close.  This bug will occur when clicking both dates, or when typing both dates - but it will only occur after the calendar has closed.
4. Re-open the calendar with the calendar icon button (bug does not happen when clicking the input fields)
5. Wait several minutes to kill the tab as the entire page ceases to function.

I've replicated this in both chrome and firefox, and I'm going to attempt to fork the repo and fix it.

Valar05:
Fix for a bug that used to cause an infinite loop.  With autoSync on and with the dates already filled, the page used to infinitely try to sync dates.  Now this will no longer happen, as there is no reason to autoSync when end date is already populated.

closes #483

### DIFF
--- a/packages/date/src/DateRange.js
+++ b/packages/date/src/DateRange.js
@@ -67,7 +67,6 @@ const DateRange = ({
   const [focusedInput, setFocusedInput] = useState(null);
 
   const calendarIconRef = useRef();
-
   const startId = `${(id || name).replace(/[^a-zA-Z0-9]/gi, '')}-start`;
 
   const endId = `${(id || name).replace(/[^a-zA-Z0-9]/gi, '')}-end`;
@@ -159,7 +158,7 @@ const DateRange = ({
 
   const onFocusChange = async input => {
     if (!input) await setFieldTouched(name, true);
-    if (autoSync) await syncDates();
+    if (autoSync && !endValue) await syncDates();
     setFocusedInput(input);
     if (onPickerFocusChange) onPickerFocusChange({ focusedInput: input });
   };
--- a/packages/date/src/DateRange.js
+++ b/packages/date/src/DateRange.js
@@ -158,9 +158,9 @@ const DateRange = ({
   };
 
   const onFocusChange = async input => {
-    if (!autoSync && !input) await setFieldTouched(name, true);
-    if (autoSync && focusedInput === input) await syncDates();
-    setFocusedInput(input);
+    if (!input) await setFieldTouched(name, true);
+    if (autoSync) await syncDates();
+    if (focusedInput !== input) setFocusedInput(input);
     if (onPickerFocusChange) onPickerFocusChange({ focusedInput: input });
   };
 
--- a/packages/date/src/DateRange.js
+++ b/packages/date/src/DateRange.js
@@ -69,6 +69,7 @@ const DateRange = ({
   const calendarIconRef = useRef();
 
   const startId = `${(id || name).replace(/[^a-zA-Z0-9]/gi, '')}-start`;
+
   const endId = `${(id || name).replace(/[^a-zA-Z0-9]/gi, '')}-end`;
 
   const startValue = get(value, startKey);
--- a/packages/date/src/DateRange.js
+++ b/packages/date/src/DateRange.js
@@ -159,7 +159,7 @@ const DateRange = ({
 
   const onFocusChange = async input => {
     if (!autoSync && !input) await setFieldTouched(name, true);
-    if (autoSync && !(focusedInput !== input)) await syncDates();
+    if (autoSync && focusedInput === input) await syncDates();
     setFocusedInput(input);
     if (onPickerFocusChange) onPickerFocusChange({ focusedInput: input });
   };
--- a/packages/date/src/DateRange.js
+++ b/packages/date/src/DateRange.js
@@ -158,8 +158,8 @@ const DateRange = ({
   };
 
   const onFocusChange = async input => {
-    if (!input) await setFieldTouched(name, true);
-    if (autoSync && !endValue) await syncDates();
+    if (!autoSync && !input) await setFieldTouched(name, true);
+    if (autoSync && !(focusedInput !== input)) await syncDates();
     setFocusedInput(input);
     if (onPickerFocusChange) onPickerFocusChange({ focusedInput: input });
   };
--- a/packages/date/src/DateRange.js
+++ b/packages/date/src/DateRange.js
@@ -67,8 +67,8 @@ const DateRange = ({
   const [focusedInput, setFocusedInput] = useState(null);
 
   const calendarIconRef = useRef();
-  const startId = `${(id || name).replace(/[^a-zA-Z0-9]/gi, '')}-start`;
 
+  const startId = `${(id || name).replace(/[^a-zA-Z0-9]/gi, '')}-start`;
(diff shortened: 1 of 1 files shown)


GoPro16 (commented):
I'm not sure what the exact expected behavior is in the below example but according to the `syncDates` method that is being used you can have an end date and then on focus change sync with the start date.

The user enters an `endValue` and that `endValue` is put into the `startValue`. [code link](https://github.com/Availity/availity-react/blob/master/packages/date/src/DateRange.js#L149)

I'm against the whole sync dates problem as you run into edge cases like this that make user land hell if they don't follow your expected pattern.

This may be a better question for @jordan-a-young  to answer or if you know what that use case should be originally.

If that is intended then it should be `autoSync && !input` meaning we unfocused all inputs not just the start date. Or we fix syncDates to only sync from the start to the end date.

packages/date/src/DateRange.js
```diff
@@ -159,7 +158,7 @@ const DateRange = ({
 
   const onFocusChange = async input => {
     if (!input) await setFieldTouched(name, true);
```
GoPro16, line 161: So the following should work if you can fix it. We need to only run `setFieldTouched` here if we don't have an input and autoSync is not enabled so `!input && !autoSync`. This will make sure we don't call validation too early before `syncDates` has finished.

Valar05:
> The user enters an `endValue` and that `endValue` is put into the `startValue`. [code link](https://github.com/Availity/availity-react/blob/master/packages/date/src/DateRange.js#L149)

While that might be the intended behavior - it doesn't actually work with the code as-is.  If you go to https://availity.github.io/availity-react/form/date/components/date-range-field/ and add autoSync, the only time autoSync actually works is when you enter a value into the startDate.  Entering one into endDate just pops up the error that "Start Date is required".

So far as I can tell, this change fixes the bug in question without changing any other functionality.

Adding `autoSync && !input` to the autoSync check had the effect of causing the autoSync method to never be called, so it didn't exactly fix the issue.

packages/date/src/DateRange.js
```diff
@@ -159,7 +158,7 @@ const DateRange = ({
...
-    if (autoSync) await syncDates();
+    if (autoSync && !endValue) await syncDates();
```
GoPro16: And then in order to run autoSync we can either be `autoSync && (!input || input === 'endDate)`

Furthermore, we only want to call `setFieldTouched` inside of `syncDates` if we are unfocusing completely. So we should pass the `input` to the `syncDates` and the conditionally check if the input exists before calling touched.

```js
if(!input) // we are unfocusing completely
   setFieldTouched(name,true)
```

GoPro16 (changes requested):

packages/date/src/DateRange.js
chrishavekost: Agree with this idea, but if we are calling `syncDates()` behind the condition `(!input && autoSync)` then I don't think we need to pass in `input` and do a second check. I don't think we should include `endDate` since this logic should work with `startDate` too. If that's currently broken like @Valar05 said, that's another issue that needs to be fixed (if `syncDates` is being kept around)

GoPro16: It somewhat works only with `datepicker` but if you type into the field itself it actually doesn't autofill the start date back for you. So I would consider this a fix to make autoSync only work if you start with the `startdate`. Anything outside of that case is going to cause more problems trying to force it to work

Valar05: I do know the reason we started looking at this fix for autoSync is that our users would get upset if we removed it - it's quite often that they want `start` and `end` dates the same, and don't want to have to pick it twice.  So I think we definitely want to keep it, even if it's a headache to deal with the edge-cases.

GoPro16: Right, but the extreme edge-cases like `I want to pick and end date first and then auto sync the start date` is outside the boundaries of what really we can support unless we want to take over `react-dates` and support that or build our own custom date framework to start making sure the validation is working as intended.

chrishavekost: One way around that problem could be using a single `<Date />` component by default and adding an option/checkbox to incorporate a `<DateRange />`. That could eliminate the need to keep `syncDates()` and code around all these edge cases and still keep users happy.

Valar05: So I tried 
`if (autoSync && (!input || input === 'endDate'))`
and it broke the autoSync from endDate to startDate (apparently that worked if you picked in the picker, not typing/pasting).  Basically same result as the code in this PR.

However, I tried:
`if (autoSync && (!startValue || !endValue)) await syncDates();`
and it allowed me to autoSync from endDate back to startDate, if I used the picker.  I figure the only time we want to autoSync is when one of the values is empty anyway, so it seems semantically pretty okay to me.

Just from poking around, it seems like `input` is always defined in the cases that were causing issues - even if the user opened the calendar with the icon, `input` was st ill set to `startDate`

chrishavekost: So without changing anything inside `syncDates()`, this snippet inside `onFocusChange()` seems to get rid of the infinite loop that was happening when the old `focusedInput` value became equal to current `input` value. Syncing works starting at either input and using the picker, syncing works when typing into `startDate`. This could be modified so that typing into `endDate` will sync to `startDate` if it doesn't have a value, but again we're just going further into edge cases.

```
    if (!input && !autoSync) await setFieldTouched(name, true);
    if (autoSync && (focusedInput === input )) await syncDates();
```

*Edited `focusedInput === input` for clarity.
I haven't looked into where yet, but some place in the code it looks like `focusedInput` falls out of sync and that's why it can sometimes be equal to `input`. I think if we find where then we could eliminate this check.

GoPro16: > *Edited focusedInput === input for clarity.
I haven't looked into where yet, but some place in the code it looks like focusedInput falls out of sync and that's why it can sometimes be equal to input. I think if we find where then we could eliminate this check.

packages/date/src/DateRange.js
```diff
@@ -158,8 +158,8 @@ const DateRange = ({
   };
 
   const onFocusChange = async input => {
-    if (!input) await setFieldTouched(name, true);
-    if (autoSync) await syncDates();
+    if (!autoSync && !input) await setFieldTouched(name, true);
+    if (autoSync && focusedInput === input) await syncDates();
     setFocusedInput(input);
```
chrishavekost: One last request and then I think we're good. It looks like `onDatesChange` is what was causing `focusedInput` to equal `input` and then cause a loop. If we change this to 
```
if (focusedInput !== input) {
      setFocusedInput(input); 
}
```
we can change the old line back to `if(autoSync) await syncDates();` and functionality should stay the same.

chrishavekost (changes requested):

packages/date/src/DateRange.js
Valar05: That does look a lot clearer - thanks for all the help!

jordan-a-young (approved):

PR #490, opened 2020-03-19
Add debounceTimeout, shouldGetPageData props to pagination package

nylon22:
The usefulness of the `debounceTimeout` prop is self-explanatory. It does add `react-use` as a dependency of the pagination package. https://bundlephobia.com/result?p=react-use@13.27.0

The usefulness of `shouldGetPageData` is a little less clear. On one hand, we could leave it up to user-land for the `items` prop to be a function that returns an empty array when page data should not be fetched. However, `shouldGetPageData` allows for the _previous_ items to be rendered when `shouldGetPageData` is false. For example, say `<Pagination />` is rendering the results of a fuzzy search, and it should only search when the length of `q` is `0` or `>= 2`. As the user types, when `q` is of length `1`, the results from when `q` was length `0` should be returned. `shouldGetPageData` allows you to achieve this easily without having to call the `usePagination` hook.

### DIFF
--- a/packages/pagination/src/Pagination.js
+++ b/packages/pagination/src/Pagination.js
@@ -4,6 +4,7 @@ import PropTypes from 'prop-types';
 import isFunction from 'lodash.isfunction';
 import isEqual from 'lodash.isequal';
 import * as avLocalStorage from '@availity/localstorage-core';
+import { useDebounce } from 'react-use';
 
 export const PaginationContext = React.createContext();
 
@@ -26,6 +27,7 @@ const Pagination = ({
   watchList,
   resetParams,
   defaultPage,
+  debounceTimeout,
 }) => {
   const ref = React.useRef();
   const [stateCurrentPage, setPage] = useState(defaultPage);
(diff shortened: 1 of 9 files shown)


packages/pagination/src/Pagination.js
```diff
@@ -93,14 +96,21 @@ const Pagination = ({
...
-    isFunction(theItems) ? null : theItems,
-    ...watchList,
-  ]);
+  useDebounce(
+    () => {
+      if (shouldGetPageData) {
+        getPageData();
+      }
+    },
+    debounceTimeout,
```
chrishavekost, line 105: ~~If `debounceTimeout` is now a variable, do we want to add that to the dependency array?~~

Since `debounceTimeout` isn't being used inside the `useEffect` then this shouldn't be passed in. Rest LGTM

chrishavekost (changes requested):

chrishavekost (approved):

GoPro16 (approved):
What are your thoughts on changing the `shouldGetPageData` to `shouldReturnPrevious`? Meaning the default would be `false`.  Just thinking that a user implementing this doesn't know what `getPageData` is so better to give a friendlier name potentially. Just a thought but otherwise LGTM.

PR #492, opened 2020-03-20
Perf/datepicker

chrishavekost:
- remove async/await from synchronous functions
- fix bug where custom `startKey` and `endKey` couldn't set `focusedInput`
- use style objects to avoid them getting recreated on every render (even if props are the same)
- refactor `relativeRanges` for fewer and faster comparisons
- batch validation updates for Date - validation runs once in a `useEffect`
- batch validation updates for DateRange - validation runs once in a `useEffect`
- memoize current start and end moments instead of computing multiple times per render
- remove `lodash.get` from package by using `value[key] || ''` instead

### DIFF
--- a/packages/date/src/Date.js
+++ b/packages/date/src/Date.js
@@ -14,6 +14,10 @@ import { isOutsideRange, limitPropType, buildYearPickerOptions } from './utils';
 
 export const isoDateFormat = 'YYYY-MM-DD';
 
+const yearPickerStyles = {
+  minWidth: '100%', // 4 digit years not wide enough to fill column
+};
+
 const AvDate = ({
   className,
   name,
@@ -133,9 +137,7 @@ const AvDate = ({
           <select
             data-testid="yearPicker"
             aria-label="year picker"
-            style={{
-              minWidth: '100%', // 4 digit years not wide enough to fill column
-            }}
+            style={yearPickerStyles}
             value={month.year()}
             onChange={e => {
               onYearSelect(month, e.target.value);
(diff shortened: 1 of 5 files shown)


nylon22 (approved):
LGTM 🚀

nylon22 (changes requested):
This test is broken: https://github.com/Availity/availity-react/blob/master/packages/date/tests/DateRange.test.js#L499

`endDate` is an empty string rather than `undefined`

packages/date/src/DateRange.js
```diff
@@ -68,18 +70,27 @@ const DateRange = ({
...
 
   const calendarIconRef = useRef();
 
   const startId = `${(id || name).replace(/[^a-zA-Z0-9]/gi, '')}-start`;
-
   const endId = `${(id || name).replace(/[^a-zA-Z0-9]/gi, '')}-end`;
 
-  const startValue = get(value, startKey);
-  const endValue = get(value, endKey);
+  const startValue = value[startKey] || '';
```
nylon22, line 82: i believe test is broken because of this change. s/b
```
const startValue = value[startKey] || undefined;
const endValue = value[endKey] || undefined;
```

packages/date/src/Date.js
```diff
@@ -49,18 +53,27 @@ const AvDate = ({
     ''
   )}-picker`;
 
+  // Should only run validation once per real change to component, instead of each time setFieldValue/Touched is called.
+  // By batching multiple calls for validation we can avoid multiple moment comparisons of the same values
+  // and stale values can be avoided without resorting to async/await: https://github.com/jaredpalmer/formik/issues/2083#issuecomment-571259235
+  useEffect(() => {
+    if (field.value || metadata.touched) {
+      validateField(name);
```
GoPro16, line 61: clever

GoPro16 (approved):

PR #513, opened 2020-04-06
refactor(date): update @availity/date tests and docs to use newest ve…

chrishavekost:
…rsion of @availity/yup

Resolves issues with storybook ci related to https://github.com/Availity/sdk-js/pull/237

### DIFF
--- a/packages/date/package.json
+++ b/packages/date/package.json
@@ -40,7 +40,6 @@
   },
   "dependencies": {
     "@availity/icon": "^0.7.5",
-    "@availity/yup": "^3.0.2",
     "classnames": "^2.2.6",
     "lodash.pick": "^4.4.0",
     "moment": "^2.24.0",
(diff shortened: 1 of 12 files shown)


packages/date/package.json
```diff
@@ -32,15 +32,15 @@
...
+    "@availity/yup": "^3.0.2",
     "formik": "^2.0.1",
     "react": "^16.3.0",
     "reactstrap": "^8.0.0 || ^7.0.0 || ^6.0.0",
     "yup": "^0.27.0"
   },
   "dependencies": {
     "@availity/icon": "^0.7.5",
-    "@availity/yup": "^2.0.0",
+    "@availity/yup": "^3.0.2",
```
GoPro16: I think this should be a `devDep`. As its already in peerDeps since we don' require it. Otherwise lets remove it from peerDeps completely.

Also this would be a breaking change since we are forcing the user to install a major version bump.

GoPro16 (changes requested):

GoPro16 (approved):

PR #514, opened 2020-04-07
fix(date): input event triggers

djejaquino:
`onChange` listener should be invoked both by picker change and input change.

Also, when setting a field value because of user input, the field should be marked as touched.


chrishavekost (approved):
LGTM. `setFieldTouched()` will immediately show an error when the user starts typing, but I think this helps in more cases than not.

chrishavekost (changes requested):
We'll probably want to add something similar to `DateRange` as well

djejaquino:
@chrishavekost moving `setFieldTouched` inside [this condition](https://github.com/Availity/availity-react/blob/3c0f33ffb8ed2751dee2ea0b0d6abb19bc217406/packages/date/src/Date.js#L79) works for me also. It prevents unwanted validation triggering when user starts typing and makes it consistent with `onPickerChange`, which is only triggered with a valid input.

chrishavekost:
I think we do want validation even when the date is invalid though, ideally it could be debounced or something so that the user isn't immediately shown an error message when they first start typing. 

I noticed an edge case in the current code (and what would happen if we moved `setFieldTouched` into that condition) where you can: 

1. start inputting a date
2. select another field 
   - date now shows validation error
3. select date field and add more, but still invalid, input
4. select another field
   - date no longer shows validation error, and will not reset itself until user inputs or picks a valid date

So in `onInputChange` I think we should be updating `metadata.touched` even when the date is invalid just to give a more consistent experience.

Thoughts? @jordan-a-young @GoPro16 what do you think too?

djejaquino:
@chrishavekost I was working on [c233103](https://github.com/Availity/availity-react/pull/514/commits/c233103553ae99744686848546f6fd9edbcc27ae) before you commented that edge case. 
After reviewers agree on this, I can definitely change it back.

About ISO format though, I believe that even if user inputs in custom format, we should invoke `onChange` with consistent format across picker/input.

chrishavekost:
@djejaquino no worries! Good idea with the `isoFormatted` change.

djejaquino:
I don't mean to be rude, but we have a deadline on a project that relies on this. Can we get priority on this one? Thanks.

chrishavekost:
@djejaquino I'll approve with the ISO format change and `setFieldTouched` moved out of that conditional

chrishavekost (approved):

PR #515, opened 2020-04-08
fix(date): added undefined check that caused empty year picker when v…

stevenviar:
…alue: 0 passed to DateRange


GoPro16 (approved):
LGTM

PR #539, opened 2020-05-06
feat(select): add filteredOrganizations api to AvOrganizationSelect

chrishavekost:
This feature will hook into the existing `AvOrganizationSelect` component and adds the ability to filter and return organizations that match a given set of permissions and resources. For example, if a payer space app is restricted to permission `A` and resource `B`, a user can pass `A` and `B` as `parameters` into `AvOrganizationSelect` and expect the dropdown to only contain authorized organizations for that user in that app, instead of all the organizations that user belongs to.

`AvOrganizationSelect` is now a custom resource, but it is exported the same way and these changes should be non-breaking for anyone upgrading but not using `avFilteredOrganizationsApi` from `sdk-js`

Related PR: https://github.com/Availity/sdk-js/pull/254

Docs coming soon™️


packages/select/custom-resources/AvOrganizationSelect.js
```diff
@@ -0,0 +1,49 @@
...
+import PropTypes from 'prop-types';
+import {
+  avFilteredOrganizationsApi,
+  avOrganizationsApi,
+} from '@availity/api-axios';
+import { ResourceSelect } from '..';
+
+const AvOrganizationSelect = ({
+  name,
+  parameters: { permissionId = [], resourceIds = [], regionId } = {},
```
nylon22: `parameters` could be a function.

packages/select/custom-resources/AvOrganizationSelect.js
```diff
@@ -0,0 +1,49 @@
...
+    // but without spreading values, there would be new array references on each render,
+    // causing new api instances to be created each time
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [...resourceIds, ...permissionId]
+  );
+
+  return (
+    <OrganizationSelect
+      name={name}
+      parameters={{ permissionId, resourceIds, regionId }}
```
nylon22: `/organizations` does not have a `resourceIds` parameter. I think it would make more sense to make `resourceIds` its own prop and not include it in the `parameters`

packages/select/custom-resources/AvOrganizationSelect.js
```diff
@@ -0,0 +1,49 @@
...
+  const OrganizationSelect = React.useMemo(
+    () =>
+      ResourceSelect.create({
+        resource:
+          resourceIds.length > 0 && permissionId.length > 0
+            ? avFilteredOrganizationsApi
+            : avOrganizationsApi,
+        labelKey: 'name',
+      }),
+    // React will not be able to statically check if values are the same
```
nylon22: I think we should memoize the dependencies, too. Inspiration: "Dependencies Lists" section from this article: https://kentcdodds.com/blog/usememo-and-usecallback

packages/select/custom-resources/AvOrganizationSelect.js
chrishavekost: `/organizations` doesn't but `/axi-user-permissions` does, `avFilteredOrganizationsApi` will parse the query string from the `postGet` of `ResourceSelect` and break it up into separate calls to `/organizations` and `/axi-user-permissions`. Is there another way to get the `resourceIds` included in the request? I tried adding them as their own prop but I may have done it wrong

GoPro16: The long term goal is hopefully to have some parameter added or a new resolver to some ms added that allows filtering orgs on resources. That being said if we make 1 unique parameter a prop here then its going to throw off the DX and logic some for both the open source code and the user implementing it.

It seems dirty to re-use the parameters and it does feel weird that its not a parameter of the organizations endpoint but rather the `API Resource` itself. Keeping it how we have it would not overcomplicate all the merging we do for the cache unique and parameters to be sent to the postGet for the time being until we have a better solution on the server side to handle resourceId filters.

nylon22: i think putting `resourceIds` in parameters throws off the DX. As a developer, the only way I know what i can send in my parameters is by checking what query parameters the api supports. But for `<AvOrganizationSelect />`, there would be this undocumented additional parameter that also isn't really a parameter because it doesn't get used in the api request like every other parameter. That's why i believe it should be separate, and documented.

chrishavekost: What would you think of modifying `ResourceSelect` to include something like `apiMethodParams` in the `postGet` request so that it is clearly separated? Or give `ResourceSelect` the ability to call api methods besides `postGet`, this way the method can be documented with the api and params can be passed together that way.

nylon22: `additionalPostGetArgs` on `<ResourceSelect />` seems like the way to go to me. `<AvOrganizationSelect />` accepts a `resourceIds` prop, and it renders a `<ResourceSelect />` with the `resourceIds` prop inside of its `additionalPostGetArgs`

chrishavekost:
@nylon22 by adding `additionalPostGetArgs` to `ResourceSelect` and merging the filtering logic into the organizations resource in `sdk-js` I was able to get rid of the added logic/complexity in `AvOrganizationSelect` and revert to how it was before.

packages/select/src/ResourceSelect.js
```diff
@@ -25,6 +25,7 @@ const ResourceSelect = ({
   waitUntilFocused,
   defaultToOnlyOption,
   shouldSearch,
+  additionalPostGetArgs,
   ...rest
 }) => {
```
GoPro16, line 30: We also are going to need to modify the `cacheUniq` logic so if the props change the select knows to bust the cache and refetch the resource if something from `additionalPostGetArgs` updates

packages/select/resources.js
```diff
@@ -1,20 +1,24 @@
...
 
 const AvProviderSelect = ResourceSelect.create({
   resource: avProvidersApi,
   labelKey: 'uiDisplayName',
   requiredParams: ['customerId'],
   watchParams: ['customerId'],
 });
+const AvOrganizationSelect = ResourceSelect.create({
+  resource: avOrganizationsApi,
+  labelKey: 'name',
```
GoPro16: I would add an arg to the `create` method with attributes that you want passed to the `postGetAttributes` prop on the `ResourceSelect`. This way the user can just passing `resourceIds` to `OrganizationSelect` and then just put that to the `additionalPostGetArgs` prop on the resource select for them as well as spreading whatever they may additionally pass at render time

GoPro16 (changes requested):

packages/select/resources.js
chrishavekost: I think that would make it easier for the user by letting them pass in a prop/props vs wrapping those props in an object, but the `ResourceSelect.create` method will already spread the defaults and any passed in props on creation. I think in this case it's better to be explicit to the user by saying "if you put values in `additionalPostGetArgs` they will be supplied to the `postGet` method". This way would also allow any future changes to resources in `sdk-js` without us needing to modify a `postGetAttributes` for a resource in `availity-react`, we could just start passing new `additionalPostGetArgs` to the `ResourceSelect` immediately. Thoughts?

packages/select/src/ResourceSelect.js
```diff
@@ -173,12 +174,20 @@ const ResourceSelect = ({
...
-            arrayFormat: 'repeat',
-            indices: false,
-            allowDots: true,
-          }),
+          // query-string does not support nesting values, convert additionalPostGetArgs
+          // to JSON string to be parsed by API
+          qs.stringify(
+            {
+              ...params,
+              additionalPostGetArgs: JSON.stringify(additionalPostGetArgs),
```
nylon22: `additionalPostGetArgs` should be the third argument to `postGet`. not part of what gets `qs.stringified`.

docs/source/form/select/components/organization-select.md
```diff
@@ -0,0 +1,47 @@
...
+    organizations: '',
+  }}
+  onSubmit={values => apiResource.submit(values)}
+  validationSchema={yup.object().shape({
+    organizations: yup.string().isRequired('This field is required.'),
+  })}
+>
+  <AvOrganizationSelect
+    id="organization"
+    name="organization"
```
nylon22: this should match the key in the `initialValues`

docs/source/form/select/components/organization-select.md
```diff
@@ -0,0 +1,47 @@
...
+    Submit
+  </Button>
+</Form>
+```
+
+## Props
+
+Extends [ResourceSelect Props](/form/select/components/resource-select/#props).
+
+### `additionalPostGetArgs?: object`
```
nylon22: I think this is a bit confusing. `additionalPostGetArgs` is a prop from `ResourceSelect`. We've already stated we're extending the `ResourceSelect` props.

We definitely need to call out how to filter by resource id, however. I think we either need to:
1. Change how this documentation is represented. Rather than stating that `AvOrgananizationSelect` has its own prop, just make a note to include `resourceIds` in the `additionalPostGetArgs`
2. Make a separate `AvOrganizationSelect` with a custom `resourceIds` prop that is a convenience helper for it being passed into `additionalPostGetArgs`.

packages/select/src/ResourceSelect.js
```diff
@@ -173,18 +175,22 @@ const ResourceSelect = ({
...
       fetch = () =>
         resource.postGet(
-          qs.stringify(params, {
-            encode: false,
-            arrayFormat: 'repeat',
-            indices: false,
-            allowDots: true,
-          }),
+          qs.stringify(
+            { ...params },
```
nylon22: i think this should probably go back to how it was before without the unnecessary spread

nylon22 (approved):

PR #541, opened 2020-05-07
Add avatar provider

stranzhay:
This is creating an `<Avatar>` provider component that enables the settings call to happen only once and then the url will be able to used throughout an app by using the `<AvatarImage>` component.

**Breaking Changes**
All apps using the current `<Avatar>` component will have to update their components if they upgrade to the new library version. 


docs/source/components/avatar/use-avatar-context.md
```diff
@@ -0,0 +1,16 @@
...
+title: useAvatarContext
+---
+
+If you are using a function component, you can subscribe to an avatar url by using this hook.
+
+## Example
+
+```jsx
+import React from 'react';
+import { useAvatarContext } from '@availity/Avatar';
```
nylon22: `Avatar` should be lowercased in package name

nylon22:
@GoPro16 will that commit be picked up as a breaking change?

GoPro16:
No it won't since there is no commit message with `BREAKING CHANGE` in the message body.

Also not sure the use case of a context provider here. Are we expecting that the avatar image will be used in more than one place in a given application?

stranzhay:
@GoPro16 sorry. I did not realize I needed to put that in commit message. I can edit the older commit and repush. I just need to know the exact format expected. The use case here is that in a message/conversation, the avatar is used each time a message shows up. That ends up calling settings and users.me each time.

GoPro16:
Okay, so instead of relying on a context provider, for DX I would look into using `react-query`. @jordan-a-young  and I have tried them out and it seems the most logical direction for API fetching/caching for REST endpoints.

Essentially it already wraps the caching based on the args you pass in so it will only fetch what is needed if its not in the local cache. 

[Bundle Size](https://bundlephobia.com/result?p=react-query@1.3.3)

[Documentation](https://github.com/tannerlinsley/react-query)

PR #548, opened 2020-05-21
feat(select): add pageAll prop

christieb7007:
- shouldFilterByAllOptions fetches all the api options and filters the results when typing on the client side by the label shown
- Added to AvRegionSelect

### DIFF
--- a/docs/source/form/select/components/resource-select.md
+++ b/docs/source/form/select/components/resource-select.md
@@ -123,6 +123,12 @@ When true, if the `resource` only returns one result the first time it is called
 
 When false or a function that returns false, the network request won't be made. Defaults to `true`.
 
+### `shouldFilterByAllOptions: boolean`
+
+When true, `resource.all()` is called to fetch all the results, and search strings will filter by the label values instead of making another network call. **This should only be used for resources with a consistently small result set and no api search params**
+
+Example: AvRegionsSelect has a limited number of results and no api search param
+
 ### `additionalPostGetArgs?: object`
 
 This object can be used to pass additional arguments to a resource's `postGet` call. These additional arguments are separate from the `parameters` that are supported by an API and may be used for filtering or other methods called inside a resource's `postGet` method. Example for the `organizations` resource that supports `additionalPostGetArgs`:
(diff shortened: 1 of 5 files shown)


packages/select/src/ResourceSelect.js
```diff
@@ -192,15 +214,23 @@ const ResourceSelect = ({
...
-          typeof getResult === 'function'
-            ? getResult.call(resource, resp.data)
-            : resp.data[getResult];
+        let items = [];
+        if (shouldFilterByAllOptions) {
+          items = resp;
+        } else {
+          items =
+            typeof getResult === 'function'
+              ? getResult.call(resource, resp?.data)
```
nylon22: this would be a breaking change for people using this package that haven't upgraded to the latest version of workflow. we're not using optional chaining anywhere else in `availity-react`. i don't think we're ready to require people to upgrade to use this package.

packages/select/src/ResourceSelect.js
```diff
@@ -163,8 +164,29 @@ const ResourceSelect = ({
       };
     }
     if (onPageChange) onPageChange(inputValue, page);
+
+    // if the UI is filtering, all the options should be present already
+    if (
+      shouldFilterByAllOptions &&
+      !hasMore &&
+      inputValue.length > 0 &&
+      (inputValue.length >= minCharsToSearch || minCharsToSearch === undefined)
```
nylon22: i think we can safely ignore `minCharsToSearch` here. `minCharsToSearch` was introduced to reduce load on our APIs. if we're filtering client side, i don't see the need.

docs/source/form/select/components/resource-select.md
```diff
@@ -123,6 +123,12 @@ When true, if the `resource` only returns one result the first time it is called
 
 When false or a function that returns false, the network request won't be made. Defaults to `true`.
 
+### `shouldFilterByAllOptions: boolean`
```
nylon22: i think the name of this prop is difficult to understand at a glance. I'm not entirely sure what to call it - maybe something like `shouldSearchClientSide` or `pageAll` ?

GoPro16: +1 for `pageAll`

nylon22:
@christieb7007 we need to add `pageAll` to the ts def for `<ResourceSelect />` and then i think we should be good to merge

nylon22 (approved):
LGTM

PR #551, opened 2020-05-27
feat(select): add permissionsIds to organizationSelect postGetArgs

christieb7007:
allows moving the `permissionId`s used for additional filtering to the `additionalPostGetArgs` where the `resourceIds` are already located. `additionalPostGetArgs.permissionIds` will be used over `parameters.permissionId`

### DIFF
--- a/docs/source/form/select/components/organization-select.md
+++ b/docs/source/form/select/components/organization-select.md
@@ -26,6 +26,7 @@ import '@availity/yup';
     parameters={{
       regionId: 'FL',
     }}
+    // permission 1111 with resource 1234 OR permission 2222 with resource 4321 is required
     permissionIds={['1111', '2222']}
     resourceIds={['1234', '4321']}
   />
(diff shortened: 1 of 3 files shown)


docs/source/form/select/components/organization-select.md
```diff
@@ -42,4 +42,4 @@ Extends [ResourceSelect Props](/form/select/components/resource-select/#props).
 
 ### `resourceIds?: string | Array<string>`
```
GoPro16, line 46: Can you make a new header for `permissionIds` here as well?

docs/source/form/select/components/organization-select.md
```diff
@@ -24,16 +24,16 @@ import '@availity/yup';
     id="organizations"
     name="organizations"
     parameters={{
-      permissionId: ['1111', '2222'],
```
chrishavekost, line 27: I think we want to keep `permissionId` here as an option too so that everyone can see it's still a usable param when not filtering orgs

packages/select/custom-resources/AvOrganizationSelect.js
```diff
@@ -8,11 +8,16 @@ const OrganizationSelect = ResourceSelect.create({
...
+  name,
+  resourceIds,
+  permissionIds,
+  ...props
+}) => {
   return (
     <OrganizationSelect
       name={name}
-      additionalPostGetArgs={resourceIds ? { resourceIds } : undefined}
+      additionalPostGetArgs={{ resourceIds, permissionIds }}
```
chrishavekost: Need to add a check for `undefined` here otherwise `additionalPostGetArgs` will always get created and users won't be able to use the normal `postGet`
`additionalPostGetArgs={resourceIds || permissionIds ? { resourceIds, permissionIds } : undefined}`

christieb7007: Added  `if (additionalPostGetArgs && additionalPostGetArgs.resourceIds)` to the `sdk-js` code, but could add this too

chrishavekost: I think it'd be helpful here too so that [cacheUniq](https://github.com/Availity/availity-react/blob/master/packages/select/src/ResourceSelect.js#L43) doesn't end up with an object containing undefined values, but I haven't tested to see if that's functionally different

packages/select/custom-resources/AvOrganizationSelect.js
```diff
@@ -22,6 +27,12 @@ AvOrganizationSelect.propTypes = {
   resourceIds: PropTypes.oneOfType([
     PropTypes.string,
     PropTypes.arrayOf(PropTypes.string),
+    PropTypes.arrayOf(PropTypes.arrayOf(PropTypes.string)),
+  ]),
+  permissionIds: PropTypes.oneOfType([
```
nylon22, line 44: for `dx`, `permissionIds` and `resourceIds` should be able to be either strings or numbers. most apps have permissions set as constants, many of them number constants. It would be a pain to have to cast your constants to strings to avoid getting warnings using this component.

nylon22:
@christieb7007 prop type changes look good. we need to update ts defs as well

PR #552, opened 2020-05-27
fix(spaces): look at first space payer id before searching all

dnoler:
Spaces can contain multiple payer ID's. However, the first is the "primary" payer ID (a given space can only have one), any additional values in the array are additional. We sometimes see this when a payer has multiple brands and the same payer ID is listed as "additional" across multiple spaces, which causes an issue when we look for branding for a particular payer by ID. This fix will check the primary payer ID's first before falling back on the additional/secondary ID's.


GoPro16 (approved):
LGTM

PR #554, opened 2020-06-02
feat(feedback): allow for overriding zindex value

### LINKED ISSUE #553: Feedback | Allow zIndex override for Modal
### Current Behavior

Currently when the `<Feedback modal />` component is rendered in modal form there are some cases where the zIndex collides with the way `reactstrap` renders modal. There is no way to override the `zIndex` prop passed to the [Modal](https://github.com/Availity/availity-react/blob/master/packages/feedback/src/FeedbackModal.js#L7)

### Desired Behavior

You should be able to pass additional props to the `Modal` for overriding the `zIndex`.

### Suggested Solution

Tack on a new prop to `formProps` or maybe even just add it directly to the `Feedback` since we know `zIndex` is only associated to the reactstrap  `Modal`

stranzhay:
I was able to add `zIndex` directly onto `Feedback` and pass down to `FeedbackModal` to allow for overriding react-strap modal values.

Closes #553

### DIFF
--- a/docs/source/components/feedback/feedback.md
+++ b/docs/source/components/feedback/feedback.md
@@ -52,6 +52,10 @@ Callback for when the feedback is submitted. It is called with the feedback obje
 
 Whether to open the `FeedbackForm` in a modal
 
+### `zIndex?: number | string`
+
+Allows for overriding the `z-index`value from react-strap `Modal`component.
+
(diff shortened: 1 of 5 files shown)


GoPro16 (approved):
LGTM

PR #556, opened 2020-06-03
feat(select): add onError prop for api errors

christieb7007:
adds prop to catch and handle any api errors


docs/source/form/select/components/resource-select.md
```diff
@@ -129,6 +129,10 @@ When true, `resource.all()` is called to fetch all the results, and search strin
 
 Example: AvRegionsSelect has a limited number of results and no api search param
 
+### `onError: (error: object) => void`
```
nylon22: let's be more explicit

```
### `onError: (error: Error) => void`
```

packages/select/typings/ResourceSelect.d.ts
```diff
@@ -45,6 +45,7 @@ export interface ResourceSelectProps<T> extends SelectFieldProps<T> {
       ) => boolean);
   additionalPostGetArgs?: object;
   pageAll?: boolean;
+  onError?: Function;
```
nylon22: let's be more explicit here too:

```
  onError?: (error: Error) => void;
```

nylon22 (approved):
LGTM 🔥

PR #568, opened 2020-06-16
feat: add scripts to check for missing dependencies and artifactory r…

chrishavekost:
…eferences

This PR adds scripts to package.json that run during the pre-push phase of a commit. These scripts should block changes if there any missing dependencies for a package or if the lockfile has references to artifactory.

### DIFF
--- a/scripts/artifactory-check.sh
+++ b/scripts/artifactory-check.sh
@@ -12,8 +12,18 @@
 # error out if something fails
 set -e
 
-if grep -R --exclude='*.sh' 'artifactory.availity' --include='yarn.lock' ./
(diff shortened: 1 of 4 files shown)


scripts/artifactory-check.sh
```diff
@@ -0,0 +1,22 @@
...
+#   - Run:
+#       npm run check:packages
+#
+
+# error out if something fails
+set -e
+
+if grep -R --exclude='*.sh' 'artifactory.availity' --include='yarn.lock' ./
+then
+  printf "\nOne of your packages contains a dependency from registry: artifactory.availity.com.\n"
```
GoPro16, line 27: Can we add a cool text art here? Like someone with their arms crossed?

GoPro16 (approved):
LGTM. 🔥 

Can you make this same PR to sdk-js and workflow?

PR #576, opened 2020-06-28
refactor(pagination): debounce resetParams effect

nylon22:
Debounces the effect run when `resetParams` changes

This is useful for when the developer wants to use `resetParams` in place of `watchList`, which is common when using `AvResourcePagination`, since changing `watchList` does not reset the current page. Using `resetParams` instead of `watchList` when paging protects against a query being made for a page that does not exist. This change allows the developer to relieve pressure off the api being called when using `resetParams` in the same manner as `watchList`.


chrishavekost (approved):
Change LGTM

PR #584, opened 2020-07-08
add enableReinitialize Form prop

pashatay:
Currently, if you pass initialValues, you can overwrite them by passing another value - the form will update its values.

This brings a problem when your data is stored for example in redux or in another component. You get this data and pass into form. Once this values are changed (in redux / another component) - your form does not rerender automatically. 

There is a nice enableReinitialize option which is false by default. You can set it to true and your form will update itself.

### DIFF
--- a/packages/form/src/Form.js
+++ b/packages/form/src/Form.js
@@ -5,6 +5,7 @@ import { Formik, Form as FForm } from 'formik';
 
 const Form = ({
   initialValues,
+  enableReinitialize,
   onSubmit,
   initialStatus,
   initialErrors,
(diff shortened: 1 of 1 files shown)


GoPro16 (approved):
Low key did not think that was a valid prop LOL. Appreciate the PR! 🚀 

FYI @nylon22  @chrishavekost  PRs from forked branches will not be able to have a PR build since we are using a PAT that is bound to our org. I think there is a way to change this but not sure.

jordan-a-young (approved):
This is the prop I didn't know I needed

pashatay:
> Low key did not think that was a valid prop LOL. Appreciate the PR! 🚀
> 
> FYI @nylon22 @chrishavekost PRs from forked branches will not be able to have a PR build since we are using a PAT that is bound to our org. I think there is a way to change this but not sure.

I didn't know either until faced an issue w my form lol

PR #585, opened 2020-07-17
refactor(upload): replace deprecated className and activeClassName props

gregmartDOTin:
className and activeClassName were deprecated. Updated to support the previous styling


nylon22:
i believe the same change is needed for `form-upload` https://github.com/Availity/availity-react/blob/master/packages/form-upload/Upload.js#L124

nylon22 (approved):
LGTM

PR #594, opened 2020-08-12
refactor(pagination): add elementType to accepted containerTag types

gregmartDOTin:
Added elementType to accepted containerTag types


chrishavekost (changes requested):
So I was able to find [this PR](https://github.com/facebook/prop-types/pull/211) detailing `ElementType` and judging from the unit tests I think we should be fine removing `React.ComponentType` from the TS definition and `PropTypes.func` from `PaginationContent.js`. It seems like `ElementType` will cover both those types and more

chrishavekost (approved):

PR #597, opened 2020-08-18
Add pageAllSearchBy prop to resourceSelect

christieb7007:
Adds more control to how the ui search filter works

### DIFF
--- a/packages/select/src/ResourceSelect.js
+++ b/packages/select/src/ResourceSelect.js
@@ -27,6 +27,7 @@ const ResourceSelect = ({
   shouldSearch,
   additionalPostGetArgs,
   pageAll,
+  pageAllSearchBy,
   ...rest
 }) => {
   const { setFieldValue } = useFormikContext();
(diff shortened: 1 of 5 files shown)


packages/select/src/ResourceSelect.js
```diff
@@ -170,6 +171,10 @@ const ResourceSelect = ({
 
     // if the UI is filtering, all the options should be present already
     if (pageAll && !hasMore && inputValue.length > 0) {
+      if (pageAllSearchBy && typeof pageAllSearchBy === 'function') {
+        const options = pageAllSearchBy(previousOptions, inputValue);
```
nylon22: Let's await this function incase the `pageAllSearchBy` provided is async

packages/select/src/ResourceSelect.js
```diff
@@ -170,6 +171,10 @@ const ResourceSelect = ({
 
     // if the UI is filtering, all the options should be present already
     if (pageAll && !hasMore && inputValue.length > 0) {
+      if (pageAllSearchBy && typeof pageAllSearchBy === 'function') {
```
nylon22, line 174: can we get a unit test for this condition?

nylon22 (changes requested):

nylon22 (approved):
LGTM

PR #613, opened 2020-10-06
fix(date): setFieldValue even when value is not valid as a date/moment

shkn-not-strrd:
Identified issue where if value was not a valid date/moment, it would never get set, instead '', which caused unexpected behavior from validation.


nylon22 (approved):
🧯

PR #614, opened 2020-10-07
fix(date): removing custom keys for date range

stranzhay:
team reported issue with date range not being able to have the values typed in. issue was due to custom keys in `DateRange`. determined that the custom keys were unneeded complexity and to remove them.


chrishavekost (changes requested):
LGTM, we should amend the commit message though so that users know this is a breaking change. A message like this will trigger the correct versioning automatically:

```
fix(date): removing custom keys for date range

BREAKING CHANGE: custom keys for start and end dates are no longer supported
```

nylon22 (approved):
LGTM 🙌

chrishavekost (approved):

PR #619, opened 2020-10-19
Feat/file delivery api improvments

stranzhay:
This is to address payer opened tickets for Upload issues regarding fileDelivery api. These changes will enable the user to pass props to do file delivery on upload and delivery on submit.


packages/form-upload/Upload.js
```diff
@@ -37,6 +58,23 @@ const Upload = ({
 
   const fieldValue = Array.isArray(field.value) ? field.value : [];
 
+  useEffect(() => {
+    if (rest.fileDeliveryOnSubmit && isSubmitting === true) {
```
chrishavekost: This is a good start :) We're gonna need to rely on more than just `isSubmitting`, I'll find a more fleshed out example with formik, but we want to make sure the form has been validated (which is happening during `isSubmitting`) and doesn't have any errors before we start submitting the uploads

packages/form-upload/Upload.js
```diff
@@ -75,7 +113,23 @@ const Upload = ({
         } else {
           upload.start();
         }
-        if (rest.onFileUpload) rest.onFileUpload(upload);
+
+        if (
```
chrishavekost: We might want to keep `if(rest.onFileUpload) rest.onFileUpload(upload)` first in case more users have custom callbacks defined. Not sure which will be the more favored use case but we can come back to it

chrishavekost (changes requested):

packages/form-upload/Upload.js
stranzhay: I do have it here underneath in an `else if` or are you thinking of making it completely separate from the condition?

packages/form-upload/Upload.js
```diff
@@ -13,6 +14,26 @@ import FilePickerBtn from './FilePickerBtn';
...
+const fileDeliveryOnUpload = (data, config, upload) => {
+  data.deliveries[0].fileURI = upload.references[0];
+  try {
+    avFilesDeliveryApi.uploadFilesDelivery(data, config);
+  } catch (error) {
+    console.warn('File delivery was not complete', error);
+  }
+};
+
+const fileDeliveryOnSubmit = (data, config, uploads) => {
```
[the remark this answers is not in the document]
stranzhay: so for these, the only reason i intentionally chose to duplicate code was the fact of when it sends on upload it is just single Upload and on submit it sends in an array. I will revisit a refactor for this in the morning.

docs/source/form/upload/upload.md
```diff
@@ -54,6 +54,14 @@ Callback called when file is uploaded. The callback is provided the `Upload` ins
 
 Callback called when file is removed. The callback is provided two arguments. 1. the updated files and 2. the id of the file that was removed
 
+### `fileDeliveryProps?: object`
+
+This prop will list all necessary information needed to complete uploads to `fileDeliveryApi`. By using this prop, the files will be delivered upon upload and not upon submission of form.
```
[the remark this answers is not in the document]
nylon22: i would recommend splitting out the required props into their own props

docs/source/form/upload/upload.md
```diff
@@ -54,6 +54,14 @@ Callback called when file is uploaded. The callback is provided the `Upload` ins
...
+
+### `fileDeliveryOnSubmit?: boolean`
+
+This prop is used in tandem with the `fileDeliveryProps` so that your files will only get delivered to `fileDeliveryApi` when the form is being submitted. **Default:** `false`
+
```
nylon22, line 78: suggesting renaming this to `deliverFileOnSubmit`

packages/form-upload/Upload.js
```diff
@@ -13,6 +14,26 @@ import FilePickerBtn from './FilePickerBtn';
 import FileList from './FileList';
 import './styles.scss';
 
+const fileDeliveryOnUpload = (data, config, upload) => {
+  data.deliveries[0].fileURI = upload.references[0];
+  try {
+    avFilesDeliveryApi.uploadFilesDelivery(data, config);
+  } catch (error) {
+    console.warn('File delivery was not complete', error);
```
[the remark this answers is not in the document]
nylon22: Right now, the developer cannot hook into the error state of the upload component. What if we use the `setFieldError` hook to set an error state when this happens?

packages/form-upload/Upload.js
```diff
@@ -58,22 +38,39 @@ const Upload = ({
...
+        setFieldError(error);
+      }
+    });
+  };
+
   useEffect(() => {
-    if (rest.fileDeliveryOnSubmit && isSubmitting === true) {
-      fileDeliveryOnSubmit(
-        rest.fileDeliveryProps,
+    if (rest.deliverFileOnSubmit && isSubmitting === true) {
```
chrishavekost: Formik has a few [phases to form submission](https://formik.org/docs/guides/form-submission#submission-phases) that we'll need to check. Mainly we want to make sure that the submission handler is finished validating, there are no errors, and the submission handler is running. That still may not be foolproof but I'll keep looking in to it

packages/form-upload/Upload.js
```diff
@@ -58,22 +38,39 @@ const Upload = ({
...
     }
-  }, [
-    fieldValue,
-    isSubmitting,
-    rest.clientId,
-    rest.customerId,
-    rest.fileDeliveryOnSubmit,
-    rest.fileDeliveryProps,
-  ]);
+  }, [isSubmitting]);
```
chrishavekost: We want to make sure all the proper useEffect hook dependencies are included here

chrishavekost (changes requested):

packages/form-upload/Upload.js
stranzhay: which ones are considered crucial here ? VSCode is recommending all of these. "React Hook useEffect has missing dependencies: 'callFileDelivery', 'fieldValue', 'fileDeliveryProps', 'rest.clientId', 'rest.customerId', and 'rest.deliverFileOnSubmit'. Either include them or remove the dependency array." TBH, I thought that the only one that would actually matter here is the `isSubmitting`

packages/form-upload/Upload.js
```diff
@@ -58,22 +38,39 @@ const Upload = ({
...
+      : undefined;
+
+  const callFileDelivery = (data, config, upload) => {
+    if (!Array.isArray(upload)) upload = [upload];
+    upload.forEach(upload => {
+      data.deliveries[0].fileURI = upload.references[0];
+      try {
+        avFilesDeliveryApi.uploadFilesDelivery(data, config);
+      } catch (error) {
+        setFieldError(error);
```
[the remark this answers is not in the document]
nylon22: A couple thoughts:

1. The signature is wrong. `setFieldError` needs the name of the input: https://formik.org/docs/api/formik#setfielderror-field-string-errormsg-string--void

2. I'm in favor of using `error` and allowing the developer to handle it how they would like to. We render an `<Input />`, but we don't display it, so no form feedback will show as a result of setting this field error. It really just sets an error state of the form, and allows the developer to hook into that error state.

packages/form-upload/Upload.js
```diff
@@ -37,12 +54,74 @@ const Upload = ({
...
+      try {
+        for (const u of upload) {
+          const data = {
+            deliveries: [
+              {
+                deliveryChannel,
+                fileURI: u.references[0],
+                metadata:
+                  typeof fileDeliveryMetadata === 'function'
+                    ? fileDeliveryMetadata(u)
```
nylon22, line 70: we should probably await this function in case what gets passed in is async. thoughts?

packages/form-upload/Upload.js
```diff
@@ -37,12 +54,74 @@ const Upload = ({
...
+      !onFileUpload &&
+      isSubmitting === true &&
+      isValidating === false &&
+      deliverFileOnSubmit &&
+      deliveryChannel &&
+      fileDeliveryMetadata
+    ) {
+      checkValidFormAndCallFileDelivery();
+    }
+    // FIXME: add all dependencies to the array and add logic to only submit upload once
```
nylon22: is this plan for this to be fixed in this or a follow up PR? what's the complication?

packages/form-upload/tests/filesDelivery.test.js
```diff
@@ -0,0 +1,227 @@
...
+    };
+
+    const upload = createTestExtendingUpload(file, done);
+    upload.start();
+
+    fireEvent.change(inputNode, fileEvent);
+    expect(inputNode.files.length).toBe(1);
+
+    fireEvent.click(getByText('Submit'));
+    await wait(() => expect(onSubmit).toHaveBeenCalled());
```
nylon22, line 175: `wait` has been deprecated in favor of `waitFor` in a future version of RTL. Let's future proof this a bit

packages/form-upload/Upload.js
chrishavekost: Ah I hadn't thought of that, so the intended use case for when `fileDeliveryMetadata` is a function is for extracting props from the `upload` and populating specific/unique props that might be needed downstream. 

On one hand I think if something async or more complicated is needed then a dev would just use their own `onFileUpload` function anyway, on the other hand it's not hard to await this. @strange1120  what do you think?

packages/form-upload/Upload.js
chrishavekost: I think when `fileDeliveryMetadata` is included as a dependency that its reference continually changes and causes the effect hook to rerun. The issue was we were seeing multiple API calls when including all the suggested dependencies, I can address this before merging, just wanted to make sure we weren't missing edge cases by excluding values from the dependency array

packages/form-upload/tests/filesDelivery.test.js
chrishavekost: Just documenting that we saw some breaking changes when switching to `waitFor` and agreed to investigate/address this in a follow up PR or an addition to https://github.com/Availity/availity-react/pull/631

thutto (approved):

PR #625, opened 2020-10-28
Fix/autofill dot notation

### LINKED ISSUE #623: Autofill does not autofill fields if field names use dot notation
### Current Behavior

Say a field has a `name` of `patient.firstName`. `autofill` is unable to autofill that field, whereas if it were just named `firstName` it would work fine

### Expected behavior

`autofill` should be able to autofill fields regardless of if they use dot notation

nylon22:
closes #623


jordan-a-young (approved):
LGTM!

PR #626, opened 2020-10-29
Feat/react query update

jordan-a-young:
This PR updates `react-query` to v2. It has a breaking change because some function signatures and return types changed.

I made sure each hook has the ability to pass in `options` that correspond to the `QueryConfig` from `react-query` which will allow users to have better control over how the hook behaves.

TypeScript types were updated as well.

### DIFF
--- a/docs/source/components/hooks/use-organizations.md
+++ b/docs/source/components/hooks/use-organizations.md
@@ -17,7 +17,11 @@ const Component = () => {
   );
 
   return (
-    <div>{isFetching ? 'Loading...' : data.data.organizations[0].name}</div>
+    <div>
+      {isFetching
+        ? 'Loading...'
+        : data?.data?.organizations?.[0]?.name || 'No organizations found'}
+    </div>
   );
 };
 ```
--- a/docs/source/components/hooks/use-organizations.md
+++ b/docs/source/components/hooks/use-organizations.md
@@ -28,9 +28,11 @@ const Component = () => {
 
 ## Props
 
-## `config: AxiosRequestConfig`
+### `config: AxiosRequestConfig`
 
 Config passed to the `getOrganizations` call from `@availity/api-axios`.
+
+> More information about this config can be found [here](https://availity.github.io/sdk-js/api/getting-started/#config-1)
 
 ### `options?: QueryConfig
 
--- a/docs/source/components/hooks/use-organizations.md
+++ b/docs/source/components/hooks/use-organizations.md
@@ -7,16 +7,29 @@ summary: Hook that returns organizations.
 
 ```jsx
 import React from 'react';
-import { useOrganizations } from '@availity/hooks';
-// ...
+import { useCurrentUser, useOrganizations } from '@availity/hooks';
+
 const Component = () => {
-  const { data, status, error } = useOrganizations();
+  const { data: user } = useCurrentUser();
+  const { data, isFetching } = useOrganizations(
+    { params: { permissionId: ['5'], userId: user?.id } },
+    { enabled: !!user?.id }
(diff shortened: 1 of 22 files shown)


packages/hooks/src/useCurrentRegion.js
```diff
@@ -11,11 +11,5 @@ async function fetchRegion() {
...
 
 export default function useCurrentRegion(options) {
-  const { data = null, isFetching, error } = useQuery(
-    ['region'],
-    fetchRegion,
-    options
-  );
-
-  return [data, isFetching, error];
+  return useQuery(['region'], fetchRegion, options);
```
chrishavekost, line 14: in the `fetchRegion` function above we want to make sure it's safe to access the response data since it may be undefined, I think we should update this and other functions like it to something like `response?.data?.regions?.[0]?.id`

packages/hooks/tests/useCurrentUser.test.js
```diff
@@ -12,27 +11,27 @@ afterEach(() => {
...
 
 describe('useCurrentUser', () => {
   test('should set error on rejected promise', async () => {
     avUserApi.me.mockRejectedValueOnce('An error occurred');
 
-    const { getByText } = render(<Component />);
+    const { getByTestId } = render(<Component />);
 
-    await waitForElement(() => getByText('"An error occurred"'));
+    await waitForElement(() => getByTestId('invalid'));
```
chrishavekost: This will fail the test if the element isn't found, but I think it's still a best practice to have an assertion here that the element exists/is not undefined

chrishavekost (changes requested):

chrishavekost:
Looks great by the way, I like that all the function signatures are the same now

packages/hooks/tests/useOrganizations.test.js
```diff
@@ -31,10 +29,11 @@ describe('useOrganizations', () => {
       'An error occurred'
     );
 
-    const { getByText } = render(<Component />);
+    const { getByTestId } = render(<Component />);
 
-    await waitForElement(() => getByText('"An error occurred"'));
+    await waitForElement(() => getByTestId('invalid'));
   });
+
```
chrishavekost: The test below on line 37 should be `async` and the `getByTestId('loading')` should be `await`ed, I think that's why the test results are varying based on when and how many tests are run. This should also fix the tests for `useCurrentRegion` and `useCurrentUser` since they have a similar non-async test

chrishavekost (changes requested):

chrishavekost:
@nylon22 @blockbat see 8d16857 for changes to `useOrganizations` test and why. I think if we can mimic this in our other tests we will get consistent results across Jest Test Runner, `yarn test packages/hooks` and `yarn test`

jordan-a-young:
@chrishavekost instead of having 3 tests and checking for loading state in one of them. I now have 2 tests that each check for loading first, and then success/error

docs/source/components/hooks/use-organizations.md
```diff
@@ -7,16 +7,33 @@ summary: Hook that returns organizations.
...
   );
 };
-// ...
 ```
+
+## Props
+
+## `config: AxiosRequestConfig`
+
+Config passed to the `getOrganizations` call from `@availity/api-axios`.
```
nylon22, line 33: it would be useful to link to this: https://availity.github.io/sdk-js/api/getting-started/#config-1

jordan-a-young: Added

chrishavekost (approved):
LGTM 🏆

PR #631, opened 2020-11-06
Feat/upgrade everything

nylon22:
Packages with breaking changes are in their own commits


jordan-a-young:
Things look good to me. It looks like you turned off the `arrowParens` rule for eslint. Can we do the same for prettier? Also, do we want to remove the `React` imports from component files with this pr too?

nylon22:
i turned off the arrow parens rule for eslint because it conflicts with prettier. (Prettier says 👍 to parens around single function arguments. Eslint says 👎 .)

I think we're going to have to keep the React imports for a while. The new jsx transform stuff where you don't have to import React only works in conjunction with a compiler so we would have to:

1. upgrade workflow to use the jsx transform
2. require the latest version of workflow for every package
3. require react >= 17 for every package
4. cut a breaking change for every package

our support costs are low for older versions of react. i don't think we're ready

robmcguinness:
@jordan-a-young https://github.com/Availity/eslint-config-availity/issues/169

jordan-a-young:
I took a different approach. I kept no arrow parens for `eslint` and disabled `prettier` so it would stop adding the parens.

chrishavekost (approved):
LGTM 🏆

PR #632, opened 2020-11-10
fix(help): add FieldHelpIcon to export

jordan-a-young:
I updated the types by adding the `FieldHelpIcon` to the exports, and I changed the type of `HelpProvider` to be a Function Component (FC). The previous type of `React.Provider` expected a value to be passed in to `HelpProvider`. The FC type only expects children to be passed in.


chrishavekost (approved):
LGTM!

PR #648, opened 2020-12-02
Fix/allow select all

gregmartDOTin:
In multi-select, fixed `allowSelectAll` to account for null values. (Also fixed a couple of failing tests in `availity/select`


packages/select/tests/Select.test.js
```diff
@@ -160,6 +160,86 @@ describe('Select', () => {
...
+        />
+        <Button type="submit">Submit</Button>
+      </Form>
+    );
+
+    await selectItem(container, getByText, 'Select all');
+
+    await fireEvent.click(getByText('Submit'));
+
+    await wait(() => {
```
chrishavekost: `wait` is already defined in this file so I think this will be fine when it gets merged..

packages/select/src/Select.js
```diff
@@ -241,7 +241,7 @@ const Select = ({
     if (allowSelectAll && attributes.isMulti) {
       if (
         [...options, ...newOptions].length > 0 &&
-        (values[name] === undefined ||
+        (!values[name] ||
```
chrishavekost: I don't think this should come up often, but it's possible to have 
```js
const values = {
  a: true,
  b: false,
  c: null,
  d: '',
}
```
and then 
`!values[a] = false`
`!values[b] = true` -> false positive
`!values[c] = true` -> desired behavior for `null`
`!values[d] = true` -> potential false positive?
`!values[e] = true` -> desired behavior for `undefined`

when really it looks like we only want `true` returned for `undefined` or `null` values. So I think we should just make the check explicit for those two values

gregmartDOTin: Good catch. This is fixed

packages/select/tests/Select.test.js
gregmartDOTin: Figured out the issue here. Should be good to go now.

chrishavekost (approved):
LGTM

PR #652, opened 2020-12-07
refactor: support grouped options

gregmartDOTin:
Support for grouped options. Previously, grouped options displayed correctly in the dropdown and selected the correct value. However, it did not return a display value. So in the case of a multi-select, it displayed and empty badge. In the case of a single-select, it displayed an empty input. This code flattens the options before searching for the value.

### DIFF
--- a/docs/source/form/select/components/select.md
+++ b/docs/source/form/select/components/select.md
@@ -16,7 +16,7 @@ import '@availity/yup';
   initialValues={{
     justTheInput: undefined,
   }}
-  onSubmit={values => window.alert(JSON.stringify(values))}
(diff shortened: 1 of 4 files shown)


hnicbaker (approved):

gregmartDOTin:
Todo:
- [x] Write tests
- [x] Write docs

chrishavekost:
LGTM. We may want to define an `OptionType` in the `Select` typings that differentiates between "normal" options and "grouped" options which will have the `type: 'group'` be required for proper functionality

gregmartDOTin:
Good call, @chrishavekost. Added the GroupedOptions type.

PR #685, opened 2021-01-15
feat(feedback): add customPlaceholder prop for aboutOptions

### LINKED ISSUE #684: Add custom Placeholder prop for aboutOptions in Feedback Component
### Current Behavior

The current placeholder text says "This is about..." in the dropdown created by the aboutOptions prop. There is not a way to override this if needed.

### Desired Behavior

I'd like to pass down a custom placeholder prop into the Feedback and FeedbackForm components to be able to change the placeholder if desired.  If nothing is passed in for this prop it stays as "This is about..."

### Suggested Solution

I would pass down something like customPlaceholder or aboutOptionsPlaceholder into the Feedback component, then into the FeedbackForm component and edit [line 159](https://github.com/Availity/availity-react/blob/master/packages/feedback/src/FeedbackForm.js#L159) to:
`placeholder={aboutOptionsPlaceholder ? aboutOptionsPlaceholder : "This is about..."}`

This would not impact anyone currently using the component. It will be an option available to anyone currently using, or planning on using, the aboutOptions prop for the Feedback component if they want to customize the placeholder text.

N/A

### Additional context

N/A

megleighc:
closes #684

### DIFF
--- a/docusaurus/docs/components/feedback/form.md
+++ b/docusaurus/docs/components/feedback/form.md
@@ -34,9 +34,9 @@ Array of Objects containing **`icon`** (String), **`description`** (String), and
 
 Array of Objects containing **`value`** (String,Number) and **`label`** (String) properties. Allows a dropdown displaying the options provided to let the user indicate what the feedback is about.
 
-#### `aboutPlaceholder?: string`
+#### `aboutPlaceholder?: node`
 
-Placeholder text for the dropdown created via the `aboutOptions` prop. If not supplied, the placeholder will default to "This is about...".
+Placeholder text for the dropdown created via the `aboutOptions` prop. **Default:** `"This is about..."`.
 
 #### `onClose?: () => void`
 
--- a/docusaurus/docs/components/feedback/form.md
+++ b/docusaurus/docs/components/feedback/form.md
@@ -34,6 +34,10 @@ Array of Objects containing **`icon`** (String), **`description`** (String), and
 
 Array of Objects containing **`value`** (String,Number) and **`label`** (String) properties. Allows a dropdown displaying the options provided to let the user indicate what the feedback is about.
 
+#### `aboutPlaceholder?: string`
+
+Placeholder text for the dropdown created via the `aboutOptions` prop. If not supplied, the placeholder will default to "This is about...".
+
 #### `onClose?: () => void`
 
 When provided, a "Close" button is rendered and `onClose` is excuted when it's clicked.
--- a/packages/feedback/src/FeedbackForm.js
+++ b/packages/feedback/src/FeedbackForm.js
@@ -32,7 +32,7 @@ const FeedbackForm = ({
   onClose,
   faceOptions,
   aboutOptions,
-  customPlaceholder,
+  aboutPlaceholder,
   onFeedbackSent,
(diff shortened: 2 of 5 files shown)


packages/feedback/src/FeedbackForm.js
```diff
@@ -32,6 +32,7 @@ const FeedbackForm = ({
   onClose,
   faceOptions,
   aboutOptions,
+  customPlaceholder,
```
nylon22: there are multiple placeholders in this component. this prop name should be more specific. perhaps `aboutPlaceholder` ?

packages/feedback/src/FeedbackForm.js
```diff
@@ -218,6 +219,7 @@ FeedbackForm.propTypes = {
       value: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
     })
   ),
+  customPlaceholder: PropTypes.string,
```
nylon22: this prop type should be `PropTypes.node` per the react select docs https://react-select.com/props

packages/feedback/src/FeedbackForm.js
```diff
@@ -156,7 +157,7 @@ const FeedbackForm = ({
                   name="feedbackApp"
                   id="about-options"
                   data-testid="about-options"
-                  placeholder="This is about..."
+                  placeholder={customPlaceholder || 'This is about...'}
```
nylon22: we can default this to `'This is about...' with ES6 default syntax on line 35,

packages/feedback/typings/FeedbackForm.d.ts
```diff
@@ -16,6 +16,7 @@ export interface FeedbackFormProps {
   onFeedbackSent?: (feedback?: { [key: string]: any }) => void;
   faceOptions?: FaceOption[];
   aboutOptions?: AboutOption[];
+  customPlaceholder?: string;
```
nylon22: we need a corresponding change to the documentation as well

nylon22 (changes requested):

packages/feedback/typings/FeedbackForm.d.ts
megleighc: Ah yes, thank you. I meant to do that.

I noticed the URL in the README for Feedback results in "Page not Found" - (https://availity.github.io/availity-react/components/feedback). 

Am I okay to change that to the specific URL: https://availity.github.io/availity-react/components/feedback/index

docusaurus/docs/components/feedback/form.md
```diff
@@ -34,6 +34,10 @@ Array of Objects containing **`icon`** (String), **`description`** (String), and
 
 Array of Objects containing **`value`** (String,Number) and **`label`** (String) properties. Allows a dropdown displaying the options provided to let the user indicate what the feedback is about.
 
+#### `aboutPlaceholder?: string`
```
nylon22: type s/b `node`

nylon22: let's use the same convention for how the default value is represented for `prompt`

nylon22 (approved):
LGTM 🎸

PR #701, opened 2021-02-24
feat: add fileDelivery callbacks

christieb7007:
add onDeliveryError and onDeliverySuccess callbacks for FileUpload file delivery cases.

### DIFF
--- a/packages/form-upload/Upload.js
+++ b/packages/form-upload/Upload.js
@@ -25,6 +25,8 @@ const Upload = ({
   customerId,
   deliverFileOnSubmit = false,
   deliveryChannel,
+  onDeliverySuccess,
+  onDeliveryError,
   disabled = false,
   feedbackClass,
   fileDeliveryMetadata,
@@ -45,6 +47,7 @@ const Upload = ({
     isValidating,
     setFieldError,
     setFieldValue,
+    setFieldTouched,
   } = useFormikContext();
   const classes = classNames(
     className,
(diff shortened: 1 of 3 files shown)


availity-bot-cicd:

https://60368e3769ed6308372751b1--condescending-leavitt-66e607.netlify.app

packages/form-upload/Upload.js
```diff
@@ -84,9 +87,15 @@ const Upload = ({
           );
         }
 
-        await Promise.all(uploadResults);
-      } catch {
+        const responses = await Promise.all(uploadResults);
+        if (typeof onDeliverySuccess === 'function') {
+          onDeliverySuccess(responses);
```
chrishavekost, line 92: I'm wondering if we should push the `onDeliverySuccess` and `onDeliveryError` callbacks onto the uploads themselves... This would keep it more in line with our example for `onFileUpload` callbacks, and then the user's function doesn't need to be concerned with looping through the response data. I also think this would take advantage of what we have in `upload-core` where each callback pushed to an upload's `onSuccess` or `onError` array gets called for us. What do you think?

https://github.com/Availity/sdk-js/blob/master/packages/upload-core/src/upload.js#L136

https://github.com/Availity/sdk-js/blob/master/packages/upload-core/src/upload.js#L402-L411

```js
const Example = () => {
  const onUpload = (upload) => {
    if (upload) {
      upload.onSuccess.push(async () => {
        // success action
      });
      upload.onError.push(() => {
        // error action
      });
    }
  };

  const onRemove = (file) => {
    // remove action
  };

  return (
    <Form initialValues={{ myFile: undefined }}>
      <Upload
        name="myFile"
        btnText="Upload a claim"
        clientId="a"
        bucketId="b"
        customerId="c"
        onFileUpload={onUpload} // <-- add file callback function
        onFileRemove={onRemove} // <-- remove file callback function
        max={1}
      />
    </Form>
  );
};
```

chrishavekost: Looked at this again, overlooked the fact that `upload.onSuccess` and `upload.onError` are for executing callbacks after uploading to `vault`. So we would either need to pass the results of the file delivery call to the next callback, or keep everything as-is

availity-bot-cicd:

https://6037dc1e7290ec1d90c9100a--condescending-leavitt-66e607.netlify.app

PR #746, opened 2021-06-22
feat: add onPasswordSubmit handler to upload

bjnewman:
Added this optional prop to the upload component so that i can use it inside of a form and prevent the nested form submission from bubbling up and submitting my parent form.

Added some tests to demo its functionality and updated the readme.

I added an isMounted check for the UploadProgressBar component because setState is passed as a callback to the upload instance in the onSuccess, onError, onProgress functions and calling setState in these callbacks will throw memory leak warnings in tests and in dev when the component unmounts (but i know this is an ugly pattern that some people don't like so i'm open to other suggestions).

### DIFF
--- a/packages/upload/FileList.d.ts
+++ b/packages/upload/FileList.d.ts
@@ -2,6 +2,7 @@ export interface FileListProps {
     files?: Array<any>;
     children?: Function;
     onRemoveFile?: Function;
+    onPasswordSubmit?: Function;
 }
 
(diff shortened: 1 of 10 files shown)


availity-bot-cicd:

https://60d21206bc28c9dcd70ffaf1--condescending-leavitt-66e607.netlify.app

jordan-a-young (approved):
Looks good to me

PR #768, opened 2021-07-13
feat(select): add selectByValue prop to autopopulate an option

christieb7007:
pass a prop to autoselect an option based on the value or specific key of the value. Useful for when you want to autopopulate the select w/o having the entire option data for `initialValues`


availity-bot-cicd:

https://60edbd88a30be60d3b4c8f49--condescending-leavitt-66e607.netlify.app

PR #770, opened 2021-07-15
feat(feedback): updated feedback components for accessibility

tannerbyers:

### DIFF
--- a/packages/feedback/src/FeedbackButton.js
+++ b/packages/feedback/src/FeedbackButton.js
@@ -16,7 +16,11 @@ const FeedbackButton = ({
   <Button
     style={btnStyles}
     color={active === icon ? 'primary' : 'light'}
-    onClick={() => onClick(icon)}
+    aria-pressed={active === icon}
+    onClick={() => {
+      onClick(icon);
+    }}
+    onKeyDown={({ keyCode }) => keyCode === 13 && onClick(icon)}
     {...rest}
   >
     <Icon data-testid="feedback-icon" name={icon} size={iconSize} />
--- a/packages/feedback/src/FeedbackForm.js
+++ b/packages/feedback/src/FeedbackForm.js
@@ -111,7 +111,6 @@ const FeedbackForm = ({
       <Form
         aria-label="Feedback Form"
         aria-describedby="feedback-form-header"
-        aria-labelledby="feedback-form-header"
         role="form"
         onKeyDown={({ keyCode }) => keyCode === 27 && onClose()}
         data-testid="feedback-form"
@@ -170,13 +169,12 @@ const FeedbackForm = ({
                   options={aboutOptions}
                 />
               )}
-              <Label id="feedback_label" for="feedback_input">
+              <Label for="feedback_input">
                 {(active && active.label) || 'Feedback? Requests? Defects?'}
               </Label>
               <Field
                 type="textarea"
                 name="feedback"
-                aria-labelledby="feedback_label"
                 placeholder={
                   (active && active.placeholder) ||
                   'Feedback? Requests? Defects?'
--- a/packages/feedback/src/FeedbackForm.js
+++ b/packages/feedback/src/FeedbackForm.js
@@ -6,6 +6,7 @@ import {
   ModalHeader,
   ModalFooter,
   FormGroup,
+  Label,
 } from 'reactstrap';
 import { avLogMessagesApi, avRegionsApi } from '@availity/api-axios';
 import { Form, Field } from '@availity/form';
@@ -88,7 +89,8 @@ const FeedbackForm = ({
 
   return sent ? (
     <ModalHeader
-      aria-live="assertive"
+      aria-modal="false"
+      id="feedback-form-header"
       tabIndex="0"
       className="d-flex justify-content-center"
       {...modalHeaderProps}
@@ -98,8 +100,10 @@ const FeedbackForm = ({
   ) : (
     <>
       <ModalHeader
-        aria-live="assertive"
+        aria-modal="false"
         id="feedback-form-header"
+        tag="h2"
+        tabIndex="0"
         {...modalHeaderProps}
       >
         {prompt || `Tell us what you think about ${name}`}
@@ -107,7 +111,9 @@ const FeedbackForm = ({
       <Form
         aria-label="Feedback Form"
         aria-describedby="feedback-form-header"
+        aria-labelledby="feedback-form-header"
         role="form"
+        onKeyDown={({ keyCode }) => keyCode === 27 && onClose()}
         data-testid="feedback-form"
         initialValues={{
           'face-options': undefined,
@@ -143,6 +149,7 @@ const FeedbackForm = ({
           <FormGroup
             size="lg"
             id="face-options"
+            role="group"
             data-testid="face-options"
             className="d-flex flex-row justify-content-between"
           >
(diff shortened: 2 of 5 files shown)


availity-bot-cicd:

https://60f07a802178bc3a1a864ed0--condescending-leavitt-66e607.netlify.app

packages/feedback/src/FeedbackButton.js
```diff
@@ -16,7 +16,11 @@ const FeedbackButton = ({
   <Button
     style={btnStyles}
     color={active === icon ? 'primary' : 'light'}
-    onClick={() => onClick(icon)}
+    aria-pressed={active === icon}
+    onClick={() => {
+      onClick(icon);
+    }}
+    onKeyDown={({ keyCode }) => keyCode === 13 && onClick(icon)}
```
jordan-a-young, line 23: Does `onKeyDown` only fire when the Button is focused or is it listening whenever the Button is on the screen?

packages/feedback/src/FeedbackForm.js
```diff
@@ -163,16 +170,20 @@ const FeedbackForm = ({
...
+                  name="feedback"
+                  placeholder={
+                    (active && active.placeholder) ||
+                    'Feedback? Requests? Defects?'
+                  }
+                  style={{ resize: 'none' }}
+                  rows="2"
+                  id="feedback_input"
+                />
+              </Label>
```
jordan-a-young, line 174: I think you should move this closing tag `</Label>` to 175 so the `Field` is not a child of `Label`

tannerbyers: The Field is a child of the Label so that when someone uses a screen reader for the Field they can see the label text is linked to the field. Want me to use aria-labelby instead?

packages/feedback/src/FeedbackButton.js
tannerbyers: It's only when the button* is focused (from what I tested and see online)

tannerbyers:
* Launching "Give Feedback" dialog creates a keyboard   trap, and focus cannot be moved out of the area with TAB and/or SHIFT+TAB   keys.


* When ARIA attributes (state, roles, and properties) are not   valid roles according to the specification or are not used correctly,   assistive technology may not function as expected.


* Smiley face feedback controls are not grouped. Provide fieldsets   for groups of form controls. The following group(s) of custom form elements   do not provide the proper ARIA attributes: - Smiley face feedback controls in   dialogs. Ensure groups of custom form controls have a role=group and an   accessible name. Add the aria-labelledby attribute with the ID of the   on-screen label as the value. [Compliant Code Example] <h5   id="feedbackLabel">Tell us what you think about the new   Authorization/Referral.</h5>. <div role="group"   aria-labelledby="feedbackLabel" class="btn-group"> ...   </div>


* Smiley face feedback controls are not consistently identified   across pages. The following controls provide the same functionality but are   presented differently across sets of pages: - Smiley face feedback controls   look different and are programmatically different across pages. There are 3   different visual looks and programmatic approaches to this component across   pages. Ensure this control is implemented with consistent identification   (labels, accessible name, icons) and look and feel across pages.





* Give   Feedback dialog v1 focus order issues. The focus order of the following   elements do not match the expected focus order: - When "Give   Feedback" non-modal dialog is launched, focus stays on triggering   button. Focus should move to either the dialog's heading or the first   interactive element. To move programmatic focus to an element outside the   page's focus order, use the JavaScript .focus() method to shift keyboard   focus to a link or form field or a div tag that has a tabindex.


* Give Feedback non-modal dialog lacks proper boundaries, name,   role - "Give Feedback" dialog boundaries are not properly   communicated and use improper role of "menu". - Ensure that the   dialog is properly labeled. Make use of role=dialog and ensure that the   dialog is named by using aria-labelledby with the ID of the dialog's heading   as the value. - Use aria-modal=false to indicate that content outside   the non-modal dialog can be interacted with while the dialog is open. -   Remove aria-live=assertive attribute within the dialog, as proper dialog   structure and focus behavior sufficient. Aria-live=assertive should be   reserved for time-sensitive or critical notifications that interrupt content   on the page.


* Smiley face feedback controls lack state info


* Give Feedback dialog heading level inaccurate

* Ensure   heading level matches the heading's visual importance/level, The following   headings have structural markup that does not accurately communicate the   heading level: - "Give Feedback" dialog has H5 or H4 heading at   top, It is recommended that a dialog's main heading be H1 or H2 to denote the   start of new content in a dialog window. [Compliant Code Example] <h2   class="modal-title">Tell us what you think about Claim   Status</h2>

* Avoid   use of placeholder values to label or explain input. A placeholder attribute   is used in place of an input label: Textarea input in the "Give   Feedback" dialog. Developers should ensure that the placeholder   attribute is not a replacement for the label element. Provide an explicit   label for this textarea input.

* Textarea   input in the "Give Feedback" dialog. Error Message Is Not   Associated.
Ensure   text and images of text provide sufficient contrast. The following text or   image of text elements do not meet the WCAG 2.0 required minimum color   contrast ratio of 4.5:1 for standard text of 18pt (24px) or 14pt (19px) if   bolded: - "Give Feedback" textarea input placeholder (Light gray   text on white)
* Ensure custom controls provide proper textual name, role, and   state information. The following custom control does not provide state   information (i.e. "expanded", "collapsed", selected,   current): - "Give Feedback" smiley face custom controls. To provide   a state for this custom control that has two states, add aria-pressed   attribute with a value of true or false. This will inform screen reader users   whether the control is selected or not. Radio buttons could also be used for   this component: https://www.w3.org/TR/wai-aria-practices/#radiobutton




* Ensure the visible text label for a control is included in the   control's accessible name. The programmatic label for the "Give   Feedback" textarea input does not match the text from the visual label.   Input has aria-label that does not match placeholder text which is being   relied on to provide label. Ensure that visible text labels match accessible   names or include any text of the visible label(s). It is highly recommended   and preferred that the accessible name matches the visible text labels as it   would make it easier for speech input technologies to locate the form fields.   - To ensure the input's label matches the accessible name, provide an   on-screen label using the LABEL element with the "for" attribute to   explicitly associate it to the input. Remove the aria-label attribute.

packages/feedback/src/FeedbackForm.js
jordan-a-young: Yes, I think that would be better

tannerbyers: updated

packages/icon/test/Icon.test.js
```diff
@@ -25,16 +25,6 @@ describe('Icon', () => {
     expect(icon).not.toBe(null);
   });
 
-  test('should render custom aria-label', () => {
```
jordan-a-young, line 28: Since the prop still exists I think we should have a test for it

packages/feedback/src/FeedbackForm.js
LauRoxx: Do you need the aria-labelledby? Since you have "for" assigned then the label and input should already be programmatically linked (the point of wrapping the label around the field was so you could skip explicitly assigning the "for" attribute and instead have it implicitly assigned, but I know that's just preference for some). iirc aria-labelledby is more so for linking elements without an explicit label, or in cases where the for attribute is difficult to apply within the label. 

Side note: doesn't this also pull from @availity/form? I think field has a label prop that you can pass in to skip this altogether.

packages/feedback/src/FeedbackForm.js
```diff
@@ -98,16 +100,20 @@ const FeedbackForm = ({
...
+        tag="h2"
+        tabIndex="0"
         {...modalHeaderProps}
       >
         {prompt || `Tell us what you think about ${name}`}
       </ModalHeader>
       <Form
         aria-label="Feedback Form"
         aria-describedby="feedback-form-header"
+        aria-labelledby="feedback-form-header"
```
LauRoxx: aria-label and aria-labelledby do the same thing, they just accomplish it in two ways. If the label is already available on the screen, aria-labelledby just points to the element. If the label is not available on screen then aria-label allows you to write it in. aria-labelledby will take precedence when multiple labels are assigned.

availity-bot-cicd:

https://60f1a7fa5458453edc381082--condescending-leavitt-66e607.netlify.app

availity-bot-cicd:

https://60f6d97f9fcaa700aafd3572--condescending-leavitt-66e607.netlify.app

jordan-a-young (approved):

PR #777, opened 2021-08-04
fix(date): dates from calendar can be focused and selected with keyboard

gk10:


availity-bot-cicd:

https://610afc677c3b1e096e8057b2--condescending-leavitt-66e607.netlify.app

chrishavekost (commented):
The previous change seems to have been related to an infinite loop bug with the `autoSync` prop. I do think we should be setting the `focusedInput` to whatever `react-dates` is telling us, since this accessibility bug seems to be coming from a mix of multiple event handlers running and us trying to update component state in quick succession, leaving us with a stale `focusedInput` that does not get updated when it should. But we should also make sure we aren't breaking anything related to `autoSync` again.

https://github.com/Availity/availity-react/pull/488

chrishavekost:
https://github.com/Availity/availity-react/pull/488#discussion_r393950663 Looks like I'm the culprit here 😅. We should definitely test the `autoSync` feature to make sure we haven't reintroduced the bug.

chrishavekost (approved):

PR #791, opened 2021-08-16
Revert "chore: update dependabot commits to only use build prefix"

chrishavekost:


availity-bot-cicd:

https://611aac87505a3609a63aa56b--condescending-leavitt-66e607.netlify.app

PR #794, opened 2021-08-20
feat(feedback): accessibility and responsiveness updates

LauRoxx:
- all placeholders replaced by labels
- reponsive up to 320px
- support toggle added to storybook
- dropdown/modal dialog ARIA attributes


availity-bot-cicd:

https://611fd64bbcc2fe69cffa7e2b--condescending-leavitt-66e607.netlify.app

availity-bot-cicd:

https://611fdb6a9d613463df3fd274--condescending-leavitt-66e607.netlify.app

packages/feedback/src/FeedbackButton.js
```diff
@@ -23,7 +23,12 @@ const FeedbackButton = ({
     onKeyDown={({ keyCode }) => keyCode === 13 && onClick(icon)}
     {...rest}
   >
-    <Icon data-testid="feedback-icon" name={icon} size={iconSize} />
+    <Icon
+      data-testid="feedback-icon"
+      name={icon}
+      size={iconSize}
+      style={{ margin: '0px' }}
```
chrishavekost: We can move this inline style object outside the component so that it is not recreated on every render or update to the button component. We are doing something similar with `btnStyles` on line 6

This is a good multi-part article explaining the cause of renders and common fixes https://medium.com/welldone-software/why-did-you-render-mr-big-pure-react-component-2a36dd86996f

packages/feedback/src/FeedbackDropdown.js
```diff
@@ -49,7 +49,13 @@ const FeedbackDropdown = ({
   );
 
   return (
-    <DropdownMenu right style={{ width: '550px', padding: 0 }}>
+    <DropdownMenu
+      right
+      style={{ width: '550px', maxWidth: '90vw', padding: 0 }}
```
chrishavekost: Same `style` comment here

packages/feedback/src/FeedbackForm.js
```diff
@@ -199,7 +195,10 @@ const FeedbackForm = ({
         <ModalFooter>
           {showSupport ? (
             <>
-              <span style={{ display: 'inline-block', margin: 0 }}>
+              <span
+                className="d-none d-md-block"
+                style={{ display: 'inline-block', margin: 0 }}
```
chrishavekost: Same `style` comment here

packages/feedback/src/SupportDropdown.js
```diff
@@ -10,7 +10,13 @@ const SupportDropdown = ({ setSupportIsActive, feedbackToggle }) => {
   const [blocking, setBlocking] = useState(null);
 
   return (
-    <DropdownMenu right style={{ width: '550px', padding: 0 }}>
+    <DropdownMenu
+      right
+      style={{ width: '550px', maxWidth: '90vw', padding: 0 }}
```
chrishavekost: Same `style` comment here

packages/feedback/src/FeedbackForm.js
```diff
@@ -33,7 +32,7 @@ const FeedbackForm = ({
   onClose,
   faceOptions,
   aboutOptions,
-  aboutPlaceholder,
+  aboutLabel,
```
chrishavekost, line 38: If we remove `aboutPlaceholder` and replace it with `aboutLabel`, this will be a breaking change for anyone updating their version of the component. Can we keep `aboutPlaceholder` and add `aboutLabel` as a hidden label? To keep the look of feedback the same, but still hit accessibility requirements? https://www.w3.org/WAI/tutorials/forms/labels/#hiding-label-text

Or show both label and placeholder. What do you all think?

packages/feedback/src/FeedbackForm.js
```diff
@@ -165,29 +166,24 @@ const FeedbackForm = ({
...
                 rows="2"
-                id="feedback_input"
               />
               {additionalComments && (
                 <Field
                   type="textarea"
                   name="additionalFeedback"
-                  placeholder="Additional Comments... (Optional)"
+                  label="Additional Comments... (Optional)"
                   style={{ resize: 'none' }}
```
chrishavekost: We should extract all these inline style objects while we're at it.

chrishavekost (changes requested):

packages/feedback/src/FeedbackForm.js
LauRoxx: ahh, I was confusing the aboutOptions with support and thought it was a newer add, doh. Personally I'm against having both placeholder and labels showing when they say the same thing. tbh, I feel like for accessibility it's better to move away from placeholders anyway (in which case I could always keep the prop and if there's a placeholder value assign it to the label). 
but if I'm keeping this one as a placeholder do all the other inputs need to have their placeholders returned?

availity-bot-cicd:

https://612020c88bf4c2163ca9832e--condescending-leavitt-66e607.netlify.app

packages/feedback/src/FeedbackForm.js
chrishavekost: If we allow separate props for label and placeholder hopefully they wouldn't both show the same value, I agree that wouldn't be a great experience.

We shouldn't keep the `aboutPlaceholder` prop and assign it to a `label` though, that would still be a breaking change and would be misleading to devs using the prop. I think we need to decide if we just introduce the new `aboutLabel` prop and have it function alongside our `aboutPlaceholder` prop, or if we remove `aboutPlaceholder` and cut a breaking change to introduce `aboutLabel`. If we keep both, all inputs would get their placeholders returned, and then have labels added.

@jordan-a-young what do you think?

jordan-a-young: I like the breaking change idea. It seems like that will be the most straightforward and be less maintenance in the long run.

availity-bot-cicd:

https://61312e6e40e8e9110aed7e07--condescending-leavitt-66e607.netlify.app

jordan-a-young (approved):
LGTM

chrishavekost (approved):
LGTM, thank you!

PR #798, opened 2021-09-02
Feat/help accessibility

adrian-santos:
1. Adding **help** to the storybook
2. Making help accessible via keyboard

Screen Shot 2021-09-03 at 10 34 59 AM

Help Link Voiceover

### DIFF
--- a/packages/help/Help.js
+++ b/packages/help/Help.js
@@ -127,7 +127,7 @@ export const FieldHelpIcon = ({ color, size, id }) => (
     onClick={() => triggerFieldHelp(id)}
     tabIndex={0}
     onKeyPress={(e) => handleKeyPress(e, id)}
-    aria-label="This link opens the help portal, press the enter key or click the icon to open"
+    aria-label="help"
   />
 );
 
--- a/packages/help/Help.js
+++ b/packages/help/Help.js
@@ -117,7 +117,7 @@ const handleKeyPress = (e, id) => {
   }
 };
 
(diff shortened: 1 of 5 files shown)


packages/help/Help.js
```diff
@@ -111,13 +111,23 @@ export const triggerFieldHelp = (id) => {
...
   <Icon
+    role="link"
     data-testid="field-help-icon"
     name="help-circle"
     size={size}
     color={color}
     onClick={() => triggerFieldHelp(id)}
+    tabIndex={0}
+    onKeyPress={(e) => handleKeyPress(e, id)}
+    aria-label="This link opens the help portal, press the enter key or click the icon to open"
```
LauRoxx: This is probably too verbose which will be frustrating for users tabbing through when there's multiple help icons on a form.

The recommendation from level access was to just leave the aria-label as "help" but make the aria-describedby the id of the corresponding label element so users can differentiate different helps icons on a page.

adrian-santos: @LauRoxx - sounds good! I'll change the PR. I'll try the aria-describedby route. thanks!

adrian-santos: @LauRoxx - just updated the PR

jordan-a-young:
I think this all looks good. Can you add a test that makes sure the `triggerFieldHelp` function is called when the user hits enter?

adrian-santos:
@jordan-a-young - just saw your comment, added a test when user hits enter

jordan-a-young (approved):

PR #806, opened 2021-09-17
Feat/accessible form groups and input labels

LauRoxx:
- Option to add FieldHelpIcon outside of label/legend for all input types with a label/legend. (needed for accessibility)
- fix FieldHelpIcon to not be aria-hidden
- addressed a few existing errors with typing

### DIFF
--- a/packages/form/src/Radio.js
+++ b/packages/form/src/Radio.js
@@ -1,5 +1,6 @@
 import React, { useState } from 'react';
 import PropTypes from 'prop-types';
+import { FieldHelpIcon } from '@availity/help';
 import { Label, Input } from 'reactstrap';
 import { v4 as uuid } from 'uuid';
 import classNames from 'classnames';
@@ -14,6 +15,7 @@ const Radio = ({
   className,
   groupClassName,
   children,
+  helpId,
   ...attributes
 }) => {
   const { value, setValue, metadata, inline } = useRadioGroup(checkValue);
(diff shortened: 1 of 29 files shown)


availity-bot-cicd:

https://6144ade93ec77541ca3e24c6--condescending-leavitt-66e607.netlify.app

packages/form/src/Radio.js
```diff
@@ -29,6 +31,12 @@ const Radio = ({
   const errorIndicated = !!metadata.touched && !!metadata.error;
   const feedbackId =
     errorIndicated && name ? `${name}-feedback`.toLowerCase() : '';
+  const labelId = `${inputId}-label`.toLowerCase();
+  const helpIcon = helpId ? (
+    <FieldHelpIcon id={helpId} labelId={labelId} />
+  ) : (
+    false
```
jordan-a-young: I think we should have the ternary return `null` instead of `false`

packages/form/typings/Checkbox.d.ts
```diff
@@ -1,12 +1,14 @@
 import * as React from 'react';
-import { InputProps } from './Input';
 
-interface CheckboxProps extends InputProps {
+interface CheckboxProps extends React.HTMLAttributes<HTMLInputElement> {
```
jordan-a-young, line 3: I feel like we should still extend `InputProps` here.  Was it causing a problem?

LauRoxx: It said it was incorrectly extended because 'value' was incompatible. It looks like the same change was made to RadioProps a long while back to so I made the same change here.

jordan-a-young: Ah ok. Let's just run with it then

availity-bot-cicd:

https://6144b54ad8369ebc7a982714--condescending-leavitt-66e607.netlify.app

jordan-a-young (approved):

PR #810, opened 2021-09-21
fix(feedback): fixed focus issue for feedback in dropdowns

tannerbyers:
Issue Reported by @chrishavekost 

I noticed an issue within the feedback package. For some use cases that render a feedback form inside a dropdown (like at the end of a list of search results) the autoFocus prop on FeedbackButton ends up automatically closing the dropdown, leaving users unable to see or select anything.

Proposed Solution:
I think autoFocus should be made into a boolean prop, FeedbackDropdown can set this value as true and others just rendering the form could set it false?

### DIFF
--- a/packages/feedback/src/Feedback.js
+++ b/packages/feedback/src/Feedback.js
@@ -46,6 +46,7 @@ const Feedback = ({
           showSupport={showSupport}
           supportZIndex={supportZIndex}
           feedbackToggle={feedbackToggle}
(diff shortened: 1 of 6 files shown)


availity-bot-cicd:

https://614a0a34a9906900c431a53c--condescending-leavitt-66e607.netlify.app

tannerbyers:
Will fix tests names and readme

PR #812, opened 2021-09-23
Refactor/package structure

jordan-a-young:
Before you keep reading I want to apologize for how much this PR involves. I didn't realize until it was too late that this should have been done in multiple PRs/passes. However, the tests are passing and storybook is working so I believe everything is in working order.

Key changes
- Updated file structure of packages. Each package now has `src`, `tests`, and `types` folders. The `index.js` and `index.d.ts` were kept at the root. 
- `types` entry added to `package.json` of each package to point to `index.d.ts`
- Cleaned up dependencies for each package. I noticed some dependencies were listed, but not used. While others did not list dependencies at all. I updated some `peerDependency` entries to accept newer versions. I also removed some dependencies from the root `package.json` which were already being included by packages
- Added `.npmignore` to each package so the tests will not be published to npm in an effort to decrease bundle size
- Updated storybook configs to use new format in v6
- Added code splitting to some storybook stories for faster load times
- Added prettier settings and ran the formatter on js files in root directory
- Regenerated `yarn.lock` in an effort to upgrade and cleanup dependencies


bjnewman (commented):
Changes look good for upload

chrishavekost (dismissed):
🚀

availity-bot-cicd:

https://6153529e83ee980af9ce2ffe--condescending-leavitt-66e607.netlify.app

chrishavekost (approved):

PR #819, opened 2021-09-30
refactor(date): add option to reformat date input in daterange

rlkeyer:
- User entered M/D/YYYY now reformats successfully (ex. 3/3/2020 -> 03/03/2020)
- Previously if M/D/YYYY was entered, it was not accepted as a valid date


jordan-a-young:
Changes look good for the DateRange. Can you add this to the normal `Date` component too? Can you also add tests where the input is in the m/d/yyyy format?

chrishavekost (approved):
LGTM, thanks!

PR #820, opened 2021-10-01
Feat/pf 2208 restyle feedback form

danielpickett:
Refactor/restyle FeedbackForm component so it can be used by itself without looking dumb.


availity-bot-cicd:

https://6156722f4f8f3d3a77eb2753--condescending-leavitt-66e607.netlify.app

package.json
```diff
@@ -39,8 +39,9 @@
     "codecov:ci": "codecov",
     "link": "lerna link",
     "clean": "lerna clean --yes",
-    "test": "jest",
-    "test:ci": "jest --runInBand --collectCoverage",
+    "test": "TZ=UTC jest",
+    "test:watch": "TZ=UTC jest --watch",
+    "test:ci": "TZ=UTC jest --runInBand --collectCoverage",
```
danielpickett: I had some tests failing of the date/time utilities. After some googling this seemed to be the easiest fix and it worked. I hope this doesn't cause any issues in CI.

danielpickett:
Sorry about the formatting changes on the date utility related files. I tried reverting them, but it looks like that formatting is part of the pre-commit hooks, so I wasn't able to remove those changes from the PR.

package.json
jordan-a-young: Is there a way we can set this in the jest config?

packages/feedback/src/SmileField.js
```diff
@@ -3,31 +3,42 @@ import PropTypes from 'prop-types';
...
       }}
+      data-testid="face-options"
     >
-      {option.description}
-    </FeedbackButton>
-  ));
+      {options.map((option, i) => (
+        <FeedbackButton
+          autoFocus={i === 0 && autoFocusFeedbackButton}
+          style={btnStyles}
```
chrishavekost: `style` here has been replaced with an always-empty object, can we add `{ flex: 1, margin: '0 2% 0 2%' }` back or is there a reason it was removed?

packages/feedback/src/SmileField.js
```diff
@@ -3,31 +3,42 @@ import PropTypes from 'prop-types';
...
-      onClick={() => {
-        setFieldValue(name, option);
-        if (onChange) {
-          onChange(option);
-        }
+  return (
+    <div
+      style={{
+        ...groupStyles,
+        gridTemplateColumns: `repeat(${options.length}, auto)`,
```
chrishavekost: Can we stick to [using Bootstrap for layout](https://getbootstrap.com/docs/4.0/layout/grid/) here? Just so that our components don't start to diverge too much from what the UI Kit uses before we decide on an upgrade path for everything

chrishavekost (changes requested):
Looks good just had two comments

packages/feedback/src/SmileField.js
danielpickett: From my brief search, Bootstrap 4 doesn't have a utility for `display: grid` or `gap`, which is really the right solution for the CSS here. It is bootstrap 4 that we're on, isn't it?

packages/feedback/src/SmileField.js
danielpickett: Good catch, I hadn't noticed that the object was empty now, I'll remove it. I took out `{ flex: 1, margin: '0 2% 0 2%' because I think percent based margins are a bad pattern, it creates spacing that is inconsistent. Also, margin on both sides of the buttons makes alignment of the edges more difficult. That's why grid gap (see comment above) is the preferred solution now. It's better that the gaps between the children be owned by the parent, and that it only be between the children, not on both sides of each child.

package.json
danielpickett: I can look into that on Monday. I'm on PTO today, but just hopped on for a minute.

availity-bot-cicd:

https://615b268c2627b2008fee4544--condescending-leavitt-66e607.netlify.app

package.json
danielpickett: @jordan-a-young  I'm looking at this now. I went master to master thinking I would make a different branch and PR for this. Problem with that is that the tests all work fine right now. This original PR was getting pushed at about 9 or 10 PM, which I think is what made the tests fail. The particular tests that were failing had to do with checking whether a two dates were x number of days apart, meaning that these tests work fine during work hours. My time zone is currently 4 hours off UTC, so the issue will only show up in the last four working hours of the day.

That's a long way round of saying, if you're alright with it, I'm gonna put this PR in without the timezone adjustments to jest, since they'll pass right now. But I'll check this issue again in the evening and put in the fixes as their own PR. Does that work for you?

danielpickett: Here's what I'm looking at for a solution: https://newbedev.com/how-do-i-set-a-timezone-in-my-jest-config

Let me know if you have any thoughts on it.

availity-bot-cicd:

https://615b2d4ac633380fbd6bb4f9--condescending-leavitt-66e607.netlify.app

availity-bot-cicd:

https://615b2eddf247c3111d656675--condescending-leavitt-66e607.netlify.app

package.json
jordan-a-young: That sounds good to me. I was wondering why you were having problems, but he hadn't heard of it before. I think the linked solution looks good too

availity-bot-cicd:

https://615c4bc6b2deb519fd9b0018--condescending-leavitt-66e607.netlify.app

chrishavekost (approved):

chrishavekost (changes requested):
I missed the `modalHeaderProps` being removed, taking another look at this

packages/feedback/src/SmileField.js
jordan-a-young: Bootstrap has columns we should be [able to use](https://getbootstrap.com/docs/4.0/layout/grid/). Either that or [flex](https://getbootstrap.com/docs/4.0/utilities/flex/).

Do you think either of those will work?

availity-bot-cicd:

https://615cc56d9d46e94dbacb62bb--condescending-leavitt-66e607.netlify.app

chrishavekost (approved):
LGTM for real this time 😄

availity-bot-cicd:

https://615da4eacae269296b115423--condescending-leavitt-66e607.netlify.app

PR #823, opened 2021-10-01
Fix/authorize loop

### LINKED ISSUE #630: useAuthorize is stuck in an infinite loop
### Current Behavior

It's in an infinite loop for the permissions check and axi-user call

### Expected behavior

call the endpoint only once

### Reproducible example

https://stackblitz.com/edit/react-gwxkhe?file=src%2FApp.js

https://codesandbox.io/s/amazing-taussig-rmglq?file=/src/App.js

### Suggested solution(s)

@jordan-a-young has a fix

Brian Overbay had a note to pass it as a string, not an array

### Your environment

```bash
```

jordan-a-young:
I've added `react-query` to the `useAuthorize` hook. `react-query` will replace the logic that was previously using `useEffect`. The `useAuthorize` hook had functions defined inside of it previously. I moved these to `utils.js` file and added tests. 

In the future I would like to break up how the permissions are fetched so this hook can leverage the `react-query` cache to grab already fetched permissions or so the app can use the permissions this hook has fetched. 

I updated the docs and storybook to account for `react-query`.

Breaking change as `react-query` is now required and I changed the value returned to an object from an array

Closes #630.

### DIFF
--- a/docusaurus/docs/components/authorize.mdx
+++ b/docusaurus/docs/components/authorize.mdx
@@ -11,14 +11,22 @@ Check user permissions to see if the current user is authorized to see your cont
 npm
 
 ```bash
-npm install @availity/authorize @availity/api-axios axios --save
+npm install @availity/authorize @availity/api-axios axios react-query --save
 ```
 
 Yarn
 
 ```bash
-yarn add @availity/authorize @availity/api-axios axios
+yarn add @availity/authorize @availity/api-axios axios react-query
 ```
+
+## Usage
+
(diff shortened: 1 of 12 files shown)


availity-bot-cicd:

https://615773d3b9cd90359b67c4cd--condescending-leavitt-66e607.netlify.app

availity-bot-cicd:

https://615c6135fda71a1161b22451--condescending-leavitt-66e607.netlify.app

PR #832, opened 2021-10-06
fix: add timezone to jest global setup

danielpickett:
`new Date()` and `moment()` were creating dates in different timezones in some of the date/time utility tests, causing them to fail when run in the evening. In my case, my timezone has a four hour offset from UTC, so if I ran tests within the last four hours before midnight, they would fail. Other times of day they would pass.

This PR sets the timezone for jest to UTC. Now the tests should be passing any time of day.

### DIFF
--- a/global-setup.js
+++ b/global-setup.js
@@ -1,0 +1,3 @@
+module.exports = async () => {
+  process.env.TZ = 'UTC';
+};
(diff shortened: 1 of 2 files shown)


availity-bot-cicd:

https://615cf37a85989e333f3079e9--condescending-leavitt-66e607.netlify.app

chrishavekost (approved):

PR #853, opened 2021-10-22
Feat/migrate stories to csf

### LINKED ISSUE #843: Update storybook to use CSF stories and TypeScript
## 🚀 Feature request

Storybook uses a new format called [CSF](https://storybook.js.org/docs/react/api/csf) as of v6. Since we are now on v6 we should move to this format. At the same time, we should use TypeScript with the stories. TypeScript is supported out of the box with Storybook.

### Current Behavior

The stories are currently using the old `storiesOf` format and JavaScript.

### Desired Behavior

Stories to be written in CSF and TypeScript so developers can work with types more easily while developing

### Suggested Solution

Convert all of the stories while following [this guide](https://storybook.js.org/docs/react/writing-stories/introduction)

Sidenote: the two story formats will not work together so the move must be done all at once

This is for developers who will be contributing to the repo as well as developers who will be using the storybook to develop their app.

bjnewman:
This should resolve https://github.com/Availity/availity-react/issues/843 and is also a start to migrating storybook to typescript, in that the files have all been converted to typescript and the imports work and the project runs and the existing propTypes have been converted to types. 

But there is still obviously a lot of work to do and also there is a good chance that i sprinkled some extra types in there that are incorrect, like i dont think you are supposed to use React.FC for functional component stories because it whines about not having a story type on FC, etc.


danielpickett (dismissed):
🎉 🎉 🎉

availity-bot-cicd:

https://617875443b098000c129493d--condescending-leavitt-66e607.netlify.app

PR #873, opened 2021-11-10
fix(select): set new colors due to accessibility issue when error cha…

tannerbyers:
We discovered that the contrast ratio for selects with errors in the dropdown does not pass on hover.

We need to update the blues to the following colors:

Hover: #B8D4FB
Focus: #85a8fc

for more info check out SHI-1550


availity-bot-cicd:

https://618c2dafc85a0c2f4331700a--condescending-leavitt-66e607.netlify.app

jordan-a-young:
Can you add these changes to the `reactstrap-validation-select` package as well?

jordan-a-young (approved):

PR #876, opened 2021-11-12
fix(pagination): fix for pf2339 sub ticket ada86 to place name & stat…

Satyampc:
…e to correct element

### DIFF
--- a/packages/pagination/src/PaginationControls.js
+++ b/packages/pagination/src/PaginationControls.js
@@ -12,14 +12,14 @@ const PaginationControls = ({ directionLinks, autoHide, pageRange, marginPages, 
   const { pageCount, currentPage, setPage } = usePagination();
 
   const createItem = (pageNumber) => (
-    <PaginationItem key={pageNumber} active={currentPage === pageNumber} data-testid={`control-page-${pageNumber}`}>
-      <PaginationLink
-        style={{ zIndex: 'auto' }}
-        onClick={() => setPage(pageNumber)}
-        type="button"
-        aria-label={`Go to page ${pageNumber}`}
-        aria-current={currentPage === pageNumber}
-      >
+    <PaginationItem
+      key={pageNumber}
+      active={currentPage === pageNumber}
+      data-testid={`control-page-${pageNumber}`}
+      aria-label={`Go to page ${pageNumber}`}
+      aria-current={currentPage === pageNumber}
+    >
+      <PaginationLink style={{ zIndex: 'auto' }} onClick={() => setPage(pageNumber)} type="button">
         {pageNumber}
       </PaginationLink>
     </PaginationItem>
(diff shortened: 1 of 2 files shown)


availity-bot-cicd:

https://618e6ccefb5fd8e9316ad7f8--condescending-leavitt-66e607.netlify.app

LauRoxx:
Wasn't the accessibility work already completed for this component?

_From Level Access:_
[Recommendation]
Move aria-label and aria-current attributes from the LI to the BUTTON elements to ensure name and state are announced.

On the "prev" and "next" buttons, only one label is needed; both aria-label and off-screen text (sr-only) are present.

[Compliant Code Example]
```
<nav class="" aria-label="pagination">
  <ul class="pagination">
    <li class="page-item disabled"><button type="button" class="page-link" aria-label="Previous">‹ Prev</button></li>
    <li class="page-item active"><button aria-label="Page 1" aria-current="true" type="button" class="page-link" style="z-index: auto;">1</button></li>
...
  </ul>
</nav>
```

packages/pagination/src/PaginationControls.js
```diff
@@ -12,14 +12,14 @@ const PaginationControls = ({ directionLinks, autoHide, pageRange, marginPages,
...
-        onClick={() => setPage(pageNumber)}
-        type="button"
-        aria-label={`Go to page ${pageNumber}`}
-        aria-current={currentPage === pageNumber}
-      >
+    <PaginationItem
+      key={pageNumber}
+      active={currentPage === pageNumber}
+      data-testid={`control-page-${pageNumber}`}
+      aria-label={`Go to page ${pageNumber}`}
```
jordan-a-young, line 19: I think this issue was already fixed and the aria attributes should be on the `PaginationLink`. The screenshot in the ticket has the aria attributes on the `li` which is what `PaginationItem` is. Where `PaginationLink` is a `button`.

Satyampc: According to the 505 compliance ticket description: I had to change the the aria-label and aria-current from button tag to li tag, which was given in the code reference in for ticket PF-2339 and the old code had none of these changes required by the ticket. Screen-Shot for ref:

image

LauRoxx: the code you're referring to is a snapshot of the code in violation when this was reported in May, not a recommendation. The code used to have the aria-label and aria-current in the `li` when it needed to be in the `button`. It has since been moved to the `button` per the Level Access recommendation I listed in my prior comment.

Level Access moved the code recommendation into that pattern link instead of listing the recommendation in the violation, so I can see how ticket can be misleading.

Satyampc: Also this bug ticket was created around end of July.

Since no changes is required, cancelling this PR.

PR #887, opened 2021-11-16
fix(select): use withAsyncPaginate instead of SelectComponent prop fo…

### LINKED ISSUE #664: creatable is not working when using async select
both AvSelectField and AvResourceSelect are not working when i am sending creatable prop as true.
passing a function names getResult which returns a promise does works in terms of async select but sending createable has no effect on noResult "mode".
Tried AvResourceSelect also, using either resource={ getResult: myAsyncFn } and just by passing getResult={myAsyncFn}.

Can anyone have a look?

stevenviar:
…r creatable AsyncPaginate selects

Closes #664

### DIFF
--- a/packages/select/src/Select.js
+++ b/packages/select/src/Select.js
@@ -5,13 +5,15 @@ import classNames from 'classnames';
 import { useField, useFormikContext } from 'formik';
 import RSelect, { components as reactSelectComponents } from 'react-select';
 import Creatable from 'react-select/creatable';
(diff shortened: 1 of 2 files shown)


jordan-a-young (approved):
Looks good. Do we need to update Storybook as well or do the stories represent this functionality already?

stevenviar:
This functionality is already in Storybook, it was just broken because of a breaking change in `react-select-async-paginate`.

PR #888, opened 2021-11-16
feat: convert packages to typescript

jordan-a-young:
This PR is for converting the "low-hanging-fruit" packages to TypeScript

(low-hanging-fruit in this context means packages with few components and/or simple logic)

- [x] analytics
- [x] app-icon
- [x] avatar
- [x] breadcrumbs
- [ ] form
- [x] help
- [ ] hooks
- [x] icon
- [x] link
- [x] list-group
- [x] list-group-item
- [ ] page-header
- [ ] payer-logo
- [ ] phone
- [x] progress
- [x] step-wizard
- [x] training-link
- [x] typography

Other updates
- added ignore for stories to `jest` coverage
- add `.ts` files to `jest` transform
- update `lerna` to ignore stories for version updates
- random `prettier` formatting on files missing it


packages/favorites/tests/FavoriteHeart.test.js
```diff
@@ -95,8 +96,8 @@ avSettingsApi.getApplication = getApplicationMock;
 describe('FavoriteHeart', () => {
   beforeEach(() => {
     global.document.createRange = () => ({
-      setStart: () => {},
-      setEnd: () => {},
+      setStart: () => ({}),
+      setEnd: () => ({}),
```
danielpickett, line 100: Is this change intentional? One is a function that does nothing and returns undefined. The other is a function that returns an empty object.

packages/help/src/Help.tsx
```diff
@@ -1,29 +1,40 @@
 import React, { createContext, useContext, useState, useEffect, useCallback } from 'react';
-import PropTypes from 'prop-types';
 import avMessages from '@availity/message-core';
 import Icon from '@availity/icon';
 
-const HelpContext = createContext();
+type THelp = {
```
danielpickett, line 5: Is this "T" prefix an old convention for type names or something? Is it something we want to keep?

danielpickett (approved):

danielpickett:
Just a note, I'm noticing that rollup runs twice when installing on the branch.

image

I noticed this before, that when I `yarn`, lerna basically runs yarn a second time. But now it's worse since we have the postinstall script running rollup. My understanding is that, within a yarn workspaces project, the `lerna bootstrap` command just runs `yarn` under the hood. So when I run `yarn` (aka `yarn install`), it installs, but then the postinstall script includes `lerna bootstrap`, which is just running `yarn` again.

This issue wasn't introduced with this PR, so I'm just leaving this as an observation.

danielpickett:
Also, I'm noticing some packages have the .stories.tsx files right next to the component .tsx files, while others have a stories sub directory. Is the intention to clean that up over the long term?

PR #890, opened 2021-11-19
fix(reactstrap-validation-date,date): replaced react-dates with avail…

tannerbyers:
This PR replaces react-dates w/ @avality/react-dates to fix the 508 issues in the date picker temporarily.


chrishavekost (approved):
LGTM

PR #907, opened 2021-12-02
feat(select): associate help and error messages to input

LauRoxx:
react-select sets the `aria-describedby` to be the placeholder element. added hidden screenreader element to placeholder to include feedback and help message in input description.
Once there is a selection the placeholder element and `aria-describedby` attribute are removed. feedback id is then added to the less supported `aria-errormessage` once the placeholder is gone.

### DIFF
--- a/packages/select/src/Select.js
+++ b/packages/select/src/Select.js
@@ -79,7 +79,7 @@ const Select = ({
   placeholder,
   ...attributes
 }) => {
-  const [{ onChange, value: fieldValue, ...field }, { touched, error }] = useField({
+  const [{ onChange, value: fieldValue, ...field }, { touched, error: fieldError }] = useField({
     name,
     validate,
   });
@@ -97,7 +97,7 @@ const Select = ({
     <>
       {placeholder || 'Select...'}
       <span className="sr-only">
-        {(touched && error) || null} {helpMessage || null}
+        {(touched && fieldError) || null} {helpMessage || null}
       </span>
     </>
   );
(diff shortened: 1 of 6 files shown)


packages/select/tests/Select.test.js
```diff
@@ -734,4 +734,92 @@ describe('Select', () => {
...
+        }}
+        onSubmit={() => {}}
+        validationSchema={singleValueSchema('singleSelect')}
+      >
+        <Select name="singleSelect" options={options} data-testid="single-select" />
+        <Button type="submit">Submit</Button>
+      </Form>
+    );
+
+    await fireEvent.click(getByText('Submit'));
```
jordan-a-young: I don't think we need `await` for the `fireEvent` functions. It should be safe to remove here and in the other places too

packages/select/src/Select.js
```diff
@@ -90,6 +92,15 @@ const Select = ({
     _cacheUniq = [_cacheUniq];
   }
 
+  attributes.placeholder = (
+    <>
+      {attributes.placeholder || 'Select...'}
+      <span className="sr-only">
+        {(touched && hasError) || null} {helpMessage || null}
```
jordan-a-young: I'm not sure what this part `{(touched && hasError) || null}` is for? Can you elaborate?

packages/select/src/Select.js
```diff
@@ -90,6 +92,15 @@ const Select = ({
     _cacheUniq = [_cacheUniq];
   }
 
+  attributes.placeholder = (
```
jordan-a-young: I'm not a fan of adding this to the `attributes` object. Instead, can you do `const placeholder = ...` and then pass that to the `react-select` component?

LauRoxx: yea, I was on the fence about it. Was trying to avoid an extra prop if I could but I guess it's unavoidable.

Unrelated, but do you know what's going on with the PR builds? I tried twice but it said it timed out.

packages/select/src/Select.js
LauRoxx: I left the name as hasError, but technically it's the error itself. if touched & error, then it adds the error contents to the hidden portion, so its timing is the same as the feedback and error styling. is there additional formatting needed to use error? I only ever saw cases where it was a string.

packages/select/src/Select.js
jordan-a-young: I want to move away from using the `attributes` spreading pattern or otherwise I think your solution is a good.

I'm not sure what's up with the builds. A few others have failed too. I'll look into it

packages/select/src/Select.js
jordan-a-young: Ohhh, I thought it was a boolean based on the name. Based on the types it should always be a string so I think we're good.

chrishavekost: can we rename the variable from `hasError` to something like `fieldError` for clarity?

chrishavekost (approved):

PR #908, opened 2021-12-02
fix(feedback): fixed shift tab closing dropdown issue

tannerbyers:
Clicking Shift tab on the buttons was closing the dropdown since it was only checking for tab on them (to close when you leave the dropdown).

### DIFF
--- a/packages/feedback/src/FeedbackForm.js
+++ b/packages/feedback/src/FeedbackForm.js
@@ -210,11 +210,11 @@ const FeedbackForm = ({
             <Button
               onClick={onClose}
               color="secondary"
-              onKeyDown={(e) => {
-                if (e.keyCode === 13) {
+              onKeyDown={(keyCode, shiftKey) => {
+                if (keyCode === 13) {
                   onClose();
                 }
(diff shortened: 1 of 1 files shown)


packages/feedback/src/FeedbackForm.js
```diff
@@ -210,11 +210,11 @@ const FeedbackForm = ({
             <Button
               onClick={onClose}
               color="secondary"
-              onKeyDown={({ keyCode }) => {
-                if (keyCode === 13) {
+              onKeyDown={(e) => {
```
jordan-a-young: instead of switching away from destructuring here, I think we should add `shiftKey` to the args.
`({ keyCode, shiftKey }) =>`

packages/feedback/src/FeedbackForm.js
```diff
@@ -224,8 +224,8 @@ const FeedbackForm = ({
           ) : null}
 
           <Button
-            onKeyDown={({ keyCode }) => {
-              if (keyCode === 9) {
+            onKeyDown={(e) => {
```
jordan-a-young: Same thing as the other comment

jordan-a-young (approved):

PR #914, opened 2021-12-06
fix(reactstrap-validation-date,date): removed date icon to fix access…

tannerbyers:
We have multiple accessibility tickets for the date icon button (SHI-1283, SHI-1282). Instead of trying to fix them, I discussed w/ Teresa on UX and verified we can just remove the button since it serves no purpose because the same functionality is present when the inputs are focused. 

This PR removes the date icon on the right, fixes the styling so that the right border is now present, and fixes a test that was using the date icon to trigger the calender flyout.

### DIFF
--- a/packages/date/src/Date.js
+++ b/packages/date/src/Date.js
@@ -4,7 +4,6 @@ import classNames from 'classnames';
 import { useField, useFormikContext } from 'formik';
 import '@availity/react-dates/initialize';
 import { SingleDatePicker } from '@availity/react-dates';
-import Icon from '@availity/icon';
 import { InputGroup, Input, Row, Col } from 'reactstrap';
 import moment from 'moment';
 import '../polyfills';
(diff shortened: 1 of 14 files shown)


packages/date/src/Date.js
```diff
@@ -181,9 +181,6 @@ const AvDate = ({
           onFocusChange={onFocusChange}
           numberOfMonths={1}
           isOutsideRange={isOutsideRange(min, max)}
-          customInputIcon={datepicker ? calendarIcon : undefined}
```
jordan-a-young, line 184: Is there other code we can also remove now that the datepicker is being removed? Like the `datepicker` prop or `calendarIcon` prop?

tannerbyers: I've gone back through and removed the unused code and imports, Thanks!

jordan-a-young (approved):

PR #917, opened 2021-12-08
feat(pagination): add option for pagination text in PaginationControl

smashleez:
We have a requirement in RCM for our pagers to show text, ie: 1-50 of 100.

With this change, you can optionally display this in the PaginationControls.

You can also pass in a function to allow you to format the text however is needed with populatePaginationText.

### DIFF
--- a/docusaurus/docs/components/pagination/controls.md
+++ b/docusaurus/docs/components/pagination/controls.md
@@ -45,3 +45,11 @@ Add bootstrap styles to list element.
 
 Allows for custom aria-label.
 
+#### `showPaginationText?: boolean`
+
+When true, this will show pagination text next to the controls describing the current page and page range. This defaults to the following: 1-50 of 100.
+
+#### `showPaginationText?: (lower: number, upper: number, total: number) => string`
+
+When `showPaginationText` is true, this function allows for customization of the pagination text that is displayed. 
+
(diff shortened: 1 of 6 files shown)


packages/pagination/src/PaginationControls.js
```diff
@@ -144,6 +158,8 @@ PaginationControls.propTypes = {
   breakLabel: PropTypes.bool,
   listClassName: PropTypes.string,
   'aria-label': PropTypes.string,
+  showPaginationText: PropTypes.string,
```
jordan-a-young: Looks like this should be `bool` instead of `string`

packages/pagination/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@availity/pagination",
-  "version": "2.12.0",
+  "version": "2.13.0",
```
jordan-a-young: Did you bump this manually? You can move this back down to `2.12.0` and the github actions will handle bumping the version and publishing when we merge the pr

smashleez: Yes, I did - lowered it back now. I will remember that going forward!

docusaurus/docs/components/pagination/controls.md
```diff
@@ -45,3 +45,11 @@ Add bootstrap styles to list element.
 
 Allows for custom aria-label.
 
+#### `showPaginationText?: boolean`
+
+When true, this will show pagination text next to the controls describing the current page and page range. This defaults to the following: 1-50 of 100.
+
+#### `showPaginationText?: (lower: number, upper: number, total: number) => string`
```
chrishavekost: I think this should be `populatePaginationText`

chrishavekost (approved):

jordan-a-young (approved):

PR #919, opened 2021-12-08
Feat/date range pass invalid dates

### LINKED ISSUE #889: Allow DateRange to pass through invalid dates
### Current Behavior

Currently the onInputChange handler for the DateRange component will not pass through invalid dates to the formik context. This makes validation logic difficult as we cannot tell an invalid input from a blank one. 

### Desired Behavior

We should be able to tell the DateRange component to pass through invalid dates to formik and let formik handle the validation.

### Suggested Solution

I pushed a branch with a prop and some sample tests as a proof of concept - https://github.com/Availity/availity-react/compare/feat/date-range-pass-invalid-dates this is still a work in progress and definitely needs further testing

DateRange users who have optional date ranges but still want to validate user inputs.

I think there is a reasonable argument that the proposed behavior should be the default/only way to handle this, i'm actually not clear on the value we get from *not* passing through invalid dates but I'm open to suggestions and wanted to minimize breaking changes.

### Additional context

None at this time.

bjnewman:
Closes #889 

Introduces some additional complexity to the form validation but either way i think this is worth the tradeoff to get validations on non-required DateRanges

### DIFF
--- a/packages/date/src/DateRange.d.ts
+++ b/packages/date/src/DateRange.d.ts
@@ -15,6 +15,7 @@ export type DateRangeProps = {
   onChange?: (dates: { startDate: string; endDate: string }) => void;
   onPickerFocusChange?: (arg: { focusedInput: 'startDate' | 'endDate' | null }) => void;
   customArrowIcon?: React.ReactNode;
+  passThroughInvalidDates: boolean;
 } & Omit<DateProps, 'datePickerProps'>;
 
 declare const DateRange: (props: DateRangeProps) => JSX.Element;
--- a/packages/date/src/DateRange.d.ts
+++ b/packages/date/src/DateRange.d.ts
@@ -15,7 +15,7 @@ export type DateRangeProps = {
   onChange?: (dates: { startDate: string; endDate: string }) => void;
   onPickerFocusChange?: (arg: { focusedInput: 'startDate' | 'endDate' | null }) => void;
   customArrowIcon?: React.ReactNode;
-  passThroughInvalidDates?: boolean;
+  allowInvalidDates?: boolean;
 } & Omit<DateProps, 'datePickerProps'>;
 
 declare const DateRange: (props: DateRangeProps) => JSX.Element;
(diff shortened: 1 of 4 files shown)


packages/date/src/DateRange.d.ts
```diff
@@ -15,6 +15,7 @@ export type DateRangeProps = {
   onChange?: (dates: { startDate: string; endDate: string }) => void;
   onPickerFocusChange?: (arg: { focusedInput: 'startDate' | 'endDate' | null }) => void;
   customArrowIcon?: React.ReactNode;
+  passThroughInvalidDates: boolean;
```
jordan-a-young: We should mark `passThroughInvalidDates` as optional here

jordan-a-young: also, what do you think about changing the prop to be called `allowInvalidDates`?

bjnewman: do you think allowInvalidDates could be confusing as to whether it relates to form submission with invalid dates? I think its a close call - but the default behavior is counter-intuitive to me so my mental model here is likely off by a bit

jordan-a-young: I think it could be confused with whether or not to allow people to even type invalid characters, eg: letters. 

The part that gives me pause is `passThrough`.  Maybe like `shouldSetInvalidDates`?

packages/date/tests/DateRange.test.js
```diff
@@ -82,6 +82,112 @@ describe('DateRange', () => {
...
+      target: {
+        value: '01-05-1997',
+      },
+    });
+
+    // simulate submission
+    fireEvent.click(getByText('Submit'));
+
+    await waitFor(() => {
+      expect(onSubmit.mock.calls[0][0]).toEqual({
```
jordan-a-young: Rather than using the `calls` array. I think it reads easier if we use the `toHaveBeenCalledWith` helper function.

```
expect(onSubmit).toHaveBeenCalledWith({ dateRange: { startDate: '', endDate: '' } })`
```

jordan-a-young (approved):

PR #929, opened 2021-12-14
feat(analytics): switch from useRef to useMemo and add to story

jordan-a-young:
In an effort to try to understand the Analytics component better I updated to Storybook story to better showcase how it works. In doing this I came across a bug where the events would be logged twice. I updated the component to use `useMemo` instead of `useRef`.  I also updated the file structure to match the direction we want to move with the TypeScript conversions


danielpickett (approved):

packages/analytics/src/Analytics.js
```diff
@@ -1,35 +1,39 @@
...
-import { AvAnalytics } from '@availity/analytics-core';
+import React, { createContext, useContext, useEffect, useMemo } from 'react';
 import PropTypes from 'prop-types';
+import { AvAnalytics } from '@availity/analytics-core';
 
 export const AnalyticsContext = createContext();
 
+export const useAnalytics = () => {
+  const ctx = useContext(AnalyticsContext);
+  if (!ctx) throw new Error('useAnalytics must be used inside of the Analytics component');
```
chrishavekost, line 9: small nitpick but can we change to `useAnalytics must be used inside of the Analytics context` for clarity?

chrishavekost (approved):

PR #937, opened 2021-12-16
Add aria describedby to date inputs for feedback

Valar05:
This change is to add aria-describedby to date inputs so that screenreaders will read error messages from them

### DIFF
--- a/packages/date/src/DateRange.js
+++ b/packages/date/src/DateRange.js
@@ -58,6 +58,7 @@ const DateRange = ({
   innerRef,
   className,
   format,
+  ariaDescribedBy,
   datepickerProps,
   'data-testid': dataTestId,
   autoSync,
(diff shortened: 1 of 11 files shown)


packages/date/src/DateRange.js
```diff
@@ -296,6 +296,7 @@ const DateRange = ({
         <DateRangePicker
           renderMonthElement={renderMonthElement}
           minimumNights={0}
+          ariaDescribedBy={`${name.toLowerCase()}-feedback`}
```
jordan-a-young: I think we should add this to the `DateRangeField` component instead. We can pass in `ariaDescribedBy={`${name.toLowerCase()}-feedback`}` to the `DateRange` component`. Then we can pass that `ariaDesribedBy` prop to the `DateRangePicker`

packages/reactstrap-validation-date/src/AvDateRange.js
```diff
@@ -515,6 +515,7 @@ class AvDateRange extends Component {
           <DateRangePicker
             disabled={attributes.disabled}
             enableOutsideDays
+            ariaDescribedBy={`${name.toLowerCase()}-feedback`}
```
jordan-a-young: Let's use the same design here as we did for `@availity/date`

jordan-a-young (approved):

PR #941, opened 2021-12-22
feat(form): focus first invalid input on submission

LauRoxx:


packages/form/src/Form.js
```diff
@@ -1,7 +1,26 @@
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import PropTypes from 'prop-types';
+import { v4 as uuid } from 'uuid';
 import { Form as RsForm } from 'reactstrap';
-import { Formik, Form as FForm } from 'formik';
+import { Formik, Form as FForm, useFormikContext } from 'formik';
+
+const FormikOnError = ({ id }) => {
```
jordan-a-young: I think we should turn this into a hook instead of a component. We can then use this hook above line 58

tannerbyers (approved):

PR #947, opened 2021-12-29
fix(feedback): add aria-label to feedback buttons

Valar05:
Add aria-label to feedback buttons so that screen readers will read them

### DIFF
--- a/packages/feedback/src/SmileField.js
+++ b/packages/feedback/src/SmileField.js
@@ -35,6 +35,7 @@ const SmileField = ({ name, options, onChange, autoFocusFeedbackButton, onClose,
           onChange(option);
         }
       }}
+      aria-label={option.label}
     >
       {option.description}
     </FeedbackButton>
--- a/packages/feedback/tests/FeedbackButton.test.js
+++ b/packages/feedback/tests/FeedbackButton.test.js
@@ -37,7 +37,13 @@ describe('Feedback', () => {
   });
 
 
-  test('should have proper ARIA attributes', () => {
+  test('should have default ARIA label', () => {
+    const { container, getByText } = render(<FeedbackButton icon="home">Home</FeedbackButton>);
+
+    expect(container.firstChild).toHaveAttribute('aria-label', 'home button');
+  });
+
+  test('should have passed ARIA label', () => {
     const { container, getByText } = render(<FeedbackButton icon="home" aria-label="What do you like?">Home</FeedbackButton>);
 
     expect(container.firstChild).toHaveAttribute('aria-label', 'What do you like?');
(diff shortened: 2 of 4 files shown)


packages/feedback/src/SmileField.js
```diff
@@ -35,6 +35,7 @@ const SmileField = ({ name, options, onChange, autoFocusFeedbackButton, onClose,
           onChange(option);
         }
       }}
+      aria-label={option.label}
```
chrishavekost, line 38: I'm not sure that each `option` is guaranteed to a have a `label`, so providing a unique default like `aria-label={ option.label || feedbackButton${i} }` may be good here.

We also want the `FeedbackButton` itself to have the `aria-label` prop on the `Button` it renders. This happens by chance now because we spread `...rest` there, but this only makes `FeedbackButton` more accessible if you use it through `SmileField`

Current code:
```js
const FeedbackButton = ({ onClick, icon, active, children, iconSize, ...rest }) => (
  <Button
    style={btnStyles}
    color={active === icon ? 'primary' : 'light'}
    aria-pressed={active === icon}
    onClick={() => {
      onClick(icon);
    }}
    onKeyDown={({ keyCode }) => keyCode === 13 && onClick(icon)}
    {...rest} // aria-label from SmileField gets spread here
  >
    <Icon data-testid="feedback-icon" name={icon} size={iconSize} style={iconStyles} />
    <span className="sr-only">{children}</span>
  </Button>
);
```

But maybe we can do something like:
```js
const FeedbackButton = ({ onClick, icon, active, children, iconSize, ariaLabel, ...rest }) => (
  <Button
    style={btnStyles}
    color={active === icon ? 'primary' : 'light'}
    aria-pressed={active === icon}
    onClick={() => {
      onClick(icon);
    }}
    onKeyDown={({ keyCode }) => keyCode === 13 && onClick(icon)}
    aria-label={ariaLabel || someDefault}
    {...rest}
  >
    <Icon data-testid="feedback-icon" name={icon} size={iconSize} style={iconStyles} />
    <span className="sr-only">{children}</span>
  </Button>
);
```

and have `SmileField` pass in an `ariaLabel` prop. What do you think?

packages/feedback/tests/FeedbackButton.test.js
```diff
@@ -35,4 +35,11 @@ describe('Feedback', () => {
     expect(container.firstChild).toHaveAttribute('aria-pressed', 'false');
     expect(getByText('Home')).toHaveClass('sr-only');
   });
+
+
+  test('should have proper ARIA attributes', () => {
```
chrishavekost: This test should have a unique name and the lint error on line 41 can be cleaned up

chrishavekost (changes requested):

packages/feedback/src/SmileField.js
LauRoxx: Why are we adding an aria-label when there's already an accessible name via the sr-only span? we don't want competing labels.

chrishavekost: Ah good call, I didn't even notice that 😄

PR #952, opened 2022-01-04
Feat/block UI pkg

jordan-a-young:
This PR migrates the [react-block-ui](https://github.com/Availity/react-block-ui) package to this repo.

Along with the migration, I removed the `redux` integrated component as well as converted the package to TypeScript.


docusaurus/docs/components/block-ui.mdx
```diff
@@ -0,0 +1,62 @@
+---
+title: Block UI
+summary: Component for blocking sections of an interface
```
nylon22: nitpick: i'd rephrase to "Component for blocking sections of a User Interface"

docusaurus/docs/components/block-ui.mdx
```diff
@@ -0,0 +1,62 @@
...
+
+#### Storybook
+
+Check out our [Storybook](https://availity.github.io/availity-react/storybook/?path=/story/components-block-ui--default) for more examples
+
+#### Props
+
+|  Prop Name     |  Types              |  Required    |  Default |  Description                                                                           |
+| -------------- | ------------------- | ------------ | -------- | -------------------------------------------------------------------------------------- |
+| blocking       | boolean             | true         |          | Set whether the component hould block its children                                     |
```
nylon22: hould s/b should

docusaurus/docs/components/block-ui.mdx
```diff
@@ -0,0 +1,62 @@
...
+yarn add @availity/block-ui
+```
+
+### Usage
+
+#### Example
+
+```jsx
+import React, { useState } from 'react';
+import BlockUi from '@availity/block-ui';
```
nylon22, line 28: should the example include importing the css as well?

nylon22 (changes requested):

docusaurus/docs/components/block-ui.mdx
jordan-a-young: I have the component import the file itself now so the user doesn't have to. I need to look into whether this has some downsides

nylon22: hmm good question. maybe @TheSharpieOne could explain why he architected it that way initially?

TheSharpieOne: It was done this way before importing CSS in JS was a common thing. Not everyone had CSS importers configured in webpack or w/e they might have been using so it was left up to the user to import it via link tag or import statement.

packages/block-ui/package.json
```diff
@@ -0,0 +1,32 @@
+{
+  "name": "@availity/block-ui",
```
TheSharpieOne, line 2: I know there is the `@availity` scope and all, but this will essentially publish a new/separate package. It's less of a migration and more of a forking at this point.

docusaurus/docs/components/block-ui.mdx
jordan-a-young: Availity apps will have `@availity/workflow` to make sure they don't break from this change. Do we think it will be a big enough problem for other users that I should have the user import the css themselves still?

packages/block-ui/package.json
jordan-a-young: That is my intention. Do you think I should update the existing repo instead? 

We have build/publish steps, Storybook, and other resources setup in `availity-react` that I think will make development easier if this is grouped with them. I also did not want to keep supporting the  `redux` component which I felt was easier to drop if I moved the code

TheSharpieOne: You can remove redux and all with a major version bump, that is fine. Everything can be achieved that you want, I would just suggest not changing the package name so it doesn't cause both to exist. Or add (via npm deprecate or w/e) that `react-block-ui` was replaced with `@availity/block-ui`.

docusaurus/docs/components/block-ui.mdx
TheSharpieOne: I don't think it's a problem anymore. We can pretty much expect users to be able to handle css imports.

packages/block-ui/package.json
jordan-a-young: I will add a note to the other package that people can use this "new" package

packages/block-ui/package.json
```diff
@@ -0,0 +1,32 @@
...
+  "version": "0.1.0",
```
chrishavekost: If we are going to change the scope of the package, let's make this a `1.0.0` release, otherwise `BREAKING CHANGE:` commits will bump minor versions. I'm 50/50 on that change though. We can move the code here to make it easier to maintain/deploy without changing the name.

If we change, definitely need to add a note and run `npm deprecate` like was mentioned. We should also provide a codemod so that internal teams and external users can upgrade easily. 

If we don't change, should update the version to match current `react-block-ui` version

jordan-a-young: I'll work on a codemod for updating the imports.

I also agree with the versioning. Should we start at v1 or v2? I was originally thinking v2, but if were changing the name then v1 will probably work

LauRoxx:
@jordan-a-young @chrishavekost @nylon22 
Hello everyone! Reviving this PR.

Just to catch everyone up since this PR is over a year old:
- Abandoning old react-block-ui package for new @availity/block-ui package (deprecation notice to be added to react-block-ui after)
- Version currently set to v1.3.3 - was set there to match react-block-ui version before the latest 508 changes were made. Did we want to stick to latest react-block-ui version or start at v2 instead?
- Pulled out codemod Jordan had originally added to this PR

bjnewman:
@LauRoxx I agree with Chris, with the name change this should be 1.0 and mark the old package as deprecated.

nylon22 (dismissed):

LauRoxx (approved):

PR #954, opened 2022-01-05
feat(date): add configurable enableOutsideDays prop for DateRange

smashleez:
This Pull Request addresses the following: 

- In the DateRange component, the enableOutsideDays is hardcoded to always be true. We have a requirement to have this set to false. Adding it to the dateRangeProps gets overridden by this hard coding. In order to maintain the same functionality, but still have the flexibility to make this configurable, I added this as a prop and have it to defaulted to true if not supplied. 


jordan-a-young (changes requested):
Changes look great! Would you mind adding this functionality for the `Date` component as well?

packages/date/src/DateRange.js
```diff
@@ -301,7 +302,7 @@ const DateRange = ({
           {...datepickerProps}
           startDate={getDateValue(startValue)}
           startDateId={startId}
-          enableOutsideDays
```
chrishavekost, line 304: I'm hesitant about directly exposing more `react-dates` props to users since we're considering moving away from the library and would want to minimize any breaking changes when migrating. Can we move `enableOustideDays` above `{...datepickerProps}` so that your override works as intended?

This solves the issue but makes it more clear that it's not necessarily a supported prop or feature going forward, if we do decide to switch underlying libraries

chrishavekost (changes requested):

packages/date/src/DateRange.js
smashleez: I am just wondering why that property was hardcoded to be that in the first place, maybe it was just an oversight. I had expected I could pass any of the available react-dates props into the dateRangePicker props and it be respected. My preference would let dateRangeProps provide speak for themselves and than any other custom properties be specific to the Availity implementation. For example, 'numberOfMonths', 'openDirection' etc are all specific to the react-dates and you cannot override them because of their position on the input. I would to put all of these above the spread of {...datepickerProps} but that might be to overkill of a change to make with this pull request.

What library is being considered now for handling dates?

smashleez:
> Changes look great! Would you mind adding this functionality for the `Date` component as well?

Considering what @chrishavekost mentioned, would it be sufficient pass this in as a datePickerProp as it is today if we want to specify as `enableOutsideDays` set to true. There is no direct hard coding of this in the Date component, while there is on the DateRange Component.

jordan-a-young:
> > Changes look great! Would you mind adding this functionality for the `Date` component as well?
> 
> Considering what @chrishavekost mentioned, would it be sufficient pass this in as a datePickerProp as it is today if we want to specify as `enableOutsideDays` set to true. There is no direct hard coding of this in the Date component, while there is on the DateRange Component.

Good point. We can leave it alone then.

packages/date/src/DateRange.js
jordan-a-young: I am assuming it was an oversight. Looks like we need to revisit how those props are spread. At a quick glance it seems like we can clean it up a bit. We can handle that in another pr though. From my understanding, it's not so much that we have a new lib in mind, but rather the `react-dates` package is not highly maintained anymore. We also want to switch away from a `moment` based lib.

jordan-a-young (approved):

packages/date/src/DateRange.js
chrishavekost: Yeah I think that was just an oversight. And like Jordan said, no particular library in mind at the moment, but for maintenance and accessibility reasons we may have to move away

packages/date/tests/DateRange.test.js
```diff
@@ -609,4 +609,54 @@ describe('DateRange', () => {
     const pickedYear = within(nextGridYearPicker).getByText(newYear);
     expect(pickedYear).toBeDefined();
   });
+
+  test('renders outside days when enableOutsideDays is enabled', async () => {
```
chrishavekost, line 613: 🥇

chrishavekost (approved):

PR #957, opened 2022-01-05
fix(dateRange): address 'cannot read properties of undefined' console error when selecting a Range in DateRange

smashleez:
Kept this as a separate branch/issue because it is separate from my other date change...hopefully doesn't annoy anyone too much....

Whenever you click on any defined preset ranges in the DateRange DatePicker, you always get a console error:

image

I initially thought this was due to some implementation issue on my end, but I also see it in the StoryBook documentation.

This PR will be to address this issue.

### DIFF
--- a/packages/date/src/DateRange.js
+++ b/packages/date/src/DateRange.js
@@ -1,4 +1,4 @@
-import React, { useEffect, useMemo, useRef, useState } from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
 import PropTypes from 'prop-types';
 import { InputGroup, Input, Button, Row, Col } from 'reactstrap';
 import { DateRangePicker } from '@availity/react-dates';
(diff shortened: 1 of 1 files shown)


packages/date/src/DateRange.js
```diff
@@ -223,7 +223,9 @@ const DateRange = ({
 
                 // Focus the calendar icon once clicked because we don't
                 // want to get back in the loop of opening the calendar
-                calendarIconRef.current.parentElement.focus();
+                if (calendarIconRef && calendarIconRef.current) {
```
smashleez: I am unsure if this calendarIconRef is even used? Preference would be to delete it all together, but I am always seeing this ref as undefined. I don't see this ref being passed to the DateRangePicker or use anywhere else. The screen does not have any adverse behavior - just a console error.

Playing it safe by null checking things...but can remove this focus call alltogether if anyone know any better than I why this was originally added.

jordan-a-young: I think we can remove the ref completely. This was missed when we removed the ability to add a calendar icon

smashleez: Love it, I was hoping so! Doing that now

jordan-a-young (approved):

PR #975, opened 2022-01-19
feat/select: Allow for customizable search term for RemoteSelect

smashleez:
We have a need to use the RemoteSelect on RCM, but our API's do not expect 'q' as the search term.

To be able to utilize the RemoteSelect, I have add a `searchTerm` prop that will allow you to specify the name of the parameter that should apply to the typed searchTerm value.

By defaulting this to 'q' this should be backwards compatible.


jordan-a-young (approved):
LGTM

PR #976, opened 2022-01-21
Fix/authorize

### LINKED ISSUE #630: useAuthorize is stuck in an infinite loop

jordan-a-young:
Closes #630 

Fixes an issue with an infinite loading bug. In the process, I converted the pkg to TypeScript, and added `react-query`. This is a breaking change because it requires the user to add a `QueryClientProvider` to their app. 

When updating the story to use `react-query` I thought the UX was better if the cache from `react-query` is refreshed each time the story loads. I also made this change to the `hooks` stories.

I also updated the `jest` config to transform `.ts` files, and aliased `@availity/mock` so we can use `msw` in unit tests

I need to add docs about the changes before merging this


danielpickett (dismissed):

packages/authorize/src/api.ts
```diff
@@ -0,0 +1,102 @@
+import { avUserPermissionsApi, avRegionsApi } from '@availity/api-axios';
+
+import type { Permission, RequestedPermissions, RequestedResources } from './types';
+
+export const getRegion = async (region?: boolean | string): Promise<string | undefined> => {
```
chrishavekost, line 5: I think we can take advantage of [useCurrentRegion](https://github.com/Availity/availity-react/blob/master/packages/hooks/src/useCurrentRegion.js) here

packages/authorize/src/api.ts
```diff
@@ -0,0 +1,102 @@
...
+  if (region === true) {
+    const resp = await avRegionsApi.getCurrentRegion();
+
+    return resp?.data?.regions?.[0]?.id;
+  }
+
+  return region || undefined;
+};
+
+export const getPermissions = async (
```
chrishavekost, line 15: This could use [usePermissions](https://github.com/Availity/availity-react/blob/master/packages/hooks/src/usePermissions.js)

packages/authorize/src/api.ts
```diff
@@ -0,0 +1,102 @@
...
+
+export const checkPermission = (
+  permission?: Permission,
+  resources?: RequestedResources,
+  organizationId?: string,
+  customerId?: string
+): boolean => {
+  if (!permission) return false;
+
+  let isAuthorizedForCustomerId = true;
```
chrishavekost, line 38: We should probably assume a user is _not_ authorized by default

packages/authorize/src/useAuthorize.ts
```diff
@@ -0,0 +1,31 @@
...
+
+const useAuthorize = (
+  permissions: RequestedPermissions,
+  parameters: Parameters = {},
+  options?: QueryOptions
+): Result => {
+  const { organizationId, customerId, region = true, resources } = parameters;
+
+  const { data: authorized = false, isLoading } = useQuery(
+    ['authorized', permissions, region, resources, organizationId, customerId],
```
chrishavekost: May want to change the first part of this key to `'useAuthorize'` instead of `'authorized'`. IMO it would make it clearer where things came from if you are using something like `queryClient.getQueryData(queryKey)` in another part of the codebase

chrishavekost (changes requested):
None of my comments are blockers to merging this. Just requesting changes in case you agree and want to change the query key for `useAuthorize`. Otherwise those other suggestions can be picked up later

PR #980, opened 2022-01-24
feat(select): added role and aria attributes to clear indicator of se…

Valar05:
Added role and aria attributes to clear indicator of select multi

### DIFF
--- a/packages/select/src/Select.js
+++ b/packages/select/src/Select.js
@@ -24,8 +24,7 @@ const components = {
   ),
   ClearIndicator: (props) => (
     <ClearIndicator {...props}>
-      <CrossIcon />
-      <span className="sr-only">Clear all selections</span>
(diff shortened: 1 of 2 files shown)


jordan-a-young (approved):

LauRoxx:
I think we may have jumped the gun on this one. We shouldn't be adding this to the tab order until it can be activated by the keyboard. We probably need to have a larger discussion around the clear options for this component anyway. Been browsing the react-select issues and although people have mentioned wanting this in the tab order for a while, the last I've seen a maintainer comment on the issue was [here](https://github.com/JedWatson/react-select/issues/2456#issuecomment-414197563). 

If we have trouble adding keyboard support to the control or don't like having the extra tabs, level access did also suggest getting rid of the clear all button altogether.

PR #987, opened 2022-02-02
docs: update resource-select watchParams prop wording to be more descriptive

clintonlunn:
I think the description for `watchParams` was lacking some details. This will add details about the `parameters` prop being a dependency.


jordan-a-young (approved):
LGTM. Thanks Clinton

PR #988, opened 2022-02-02
fix(date): prevent loss of onChange handler when date value is manually cleared

smashleez:
This issue addresses the following with the @availity/date date component:

**Current Issue:**
When there is an onChange handler provided for either `Date` or `DateField`, that onChange handler does not get called if the value is null. For example, I can select a value with the date picker. If I backspace and delete some of that date manually, it sets the value to null (even if there is only a "partial" date left after backspacing). This creates problems when we have to report form validity to a parent component, such as a modal where the submit button is not nested within the `<Form/>`.  We have rendered that Date Input invalid by backspacing, yet we can't report on that since the onChange handler call gets skipped. We should be calling onChange handlers when the input changes and just let it flow through like any other input would do.

### DIFF
--- a/packages/date/src/Date.js
+++ b/packages/date/src/Date.js
@@ -74,15 +74,15 @@ const AvDate = ({
   };
 
   const onPickerChange = (value) => {
-    if (value === null) return;
-
     let val = value;
-    if (val instanceof Object && val.isValid()) {
-      val = val.format(isoDateFormat);
-    }
+    if (val !== null) {
+      if (val instanceof Object && val.isValid()) {
+        val = val.format(isoDateFormat);
+      }
 
-    setFieldValue(name, val, false);
-    setFieldTouched(name, true, false);
+      setFieldValue(name, val, false);
+      setFieldTouched(name, true, false);
+    }
 
     if (onChange) {
       onChange(val);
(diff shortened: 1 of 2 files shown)


chrishavekost (approved):
LGTM, we should still run a user's `onChange` function if it exists.

I do want to note though that it sounds like you're using this to run validation a second time so that you can use that result to report validity to a parent outside of Formik's context. If that's the case I think you may want to consider raising up the Formik state, or just extracting the `errors` object from Formik context so you can reuse that. Just wary of re-running validation in your own function and potentially introducing performance issues or bugs

smashleez:
> LGTM, we should still run a user's `onChange` function if it exists.
> 
> I do want to note though that it sounds like you're using this to run validation a second time so that you can use that result to report validity to a parent outside of Formik's context. If that's the case I think you may want to consider raising up the Formik state, or just extracting the `errors` object from Formik context so you can reuse that. Just wary of re-running validation in your own function and potentially introducing performance issues or bugs

I have tried a few different things similar to what you mention, and all of which are more convoluted than just fixing this bug :) 
The form itself is in a child component of another component, and passing formRefs was just getting extremely messy. I am also not running validation twice, I am trying to report on validity whenever there is a change. Thanks for the advice, though.

PR #989, opened 2022-02-03
fix(date): fix DateRangePicker props on element definition so that they are customizable and not hardcoded

smashleez:
We keep running into issues where we are trying to set a datePickerProps  and they don't get respected because it is hard-coded later down the element (after the {...datePickerProps} spread).

This PR adds more properties (autoComplete, numberOfMonths) before the spread of `{...datePickerProps}` so that we can set those and them actually propagate down to the the DateRangePicker and SingleDatePicker. We had to do this same thing before 'enableOutsideDays' and are going through and fixing some of our bugs and noticed these other ones that we need to override.

### DIFF
--- a/packages/date/src/Date.js
+++ b/packages/date/src/Date.js
@@ -167,6 +167,8 @@ const AvDate = ({
       >
         <SingleDatePicker
           renderMonthElement={renderMonthElement}
+          autoComplete="date"
+          numberOfMonths={1}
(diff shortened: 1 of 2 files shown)


jordan-a-young (approved):
LGTM. I know for some props we do this to try to maintain accessibility or uniformity across apps. These seem safe to move to me.

PR #994, opened 2022-02-09
refactor(spaces): move click area to title

gregmartDOTin:
this allows for multiple click areas on a tile. e.g. the favorties heart and the title

### DIFF
--- a/packages/spaces/src/SpacesLink.js
+++ b/packages/spaces/src/SpacesLink.js
@@ -192,14 +192,7 @@ const Link = ({
       className={classNames('spaces-hook-link', className, `spaces-${linkStyle}-link`, {
         'p-2': linkStyle === 'default',
       })}
-      tabIndex={0}
-      style={{
-        ...style,
-        cursor: link?.url ? 'pointer' : 'not-allowed',
-      }}
-      {...props}
-      {...rest}
-      aria-label={name}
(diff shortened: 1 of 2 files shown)


packages/spaces/src/SpacesLink.js
```diff
@@ -221,6 +214,13 @@ const Link = ({
                     'pt-3': stacked,
                     'text-center': stacked,
                   })}
+                  tabIndex={0}
+                  style={{
+                    cursor: link?.url ? 'pointer' : 'not-allowed',
+                  }}
+                  {...props}
```
chrishavekost, line 230: I don't think every type of `SpacesLink` will necessarily have a `body`, do you know if it's safe to be spreading `props` and `rest` here?

gregmartDOTin: It is safe to spread `props`. `rest` is not though. I moved that back. I also added all the click related stuff to the children, if not passed in as a function.

packages/spaces/styles.scss
```diff
@@ -0,0 +1,9 @@
+*[role='link'] {
```
gregmartDOTin, line 1: This feels _a little_ dangerous. But in theory, if the user is using the `link` role correctly, `role='link'` should pretty much always be treated like a `<a/>`, right?

PR #996, opened 2022-02-11
fix(table): Add ability to designate hidden columns 

smashleez:
This PR cleans up some of the previous PR comments, as well as adding the ability to designate hidden columns just by setting it as the default configuration.


jordan-a-young (approved):
Looks great!

PR #1000, opened 2022-02-16
Added aria-describedby attribute to AvSelectField

PavanGM23:

### DIFF
--- a/packages/form/src/Field.js
+++ b/packages/form/src/Field.js
@@ -3,11 +3,11 @@ import PropTypes from 'prop-types';
 import { Col, FormText, Input as RsInput, InputGroup, InputGroupAddon, InputGroupText } from 'reactstrap';
 import { v4 as uuid } from 'uuid';
 
+import { useField } from 'formik';
 import Feedback from './Feedback';
 import FormGroup from './FormGroup';
 import Input from './Input';
 import Label from './Label';
-import { useField } from 'formik';
 
 const colSizes = ['xs', 'sm', 'md', 'lg', 'xl'];
 
--- a/packages/form/src/Field.js
+++ b/packages/form/src/Field.js
@@ -57,6 +57,7 @@ const Field = ({
       readOnly={readOnly}
       feedback
       help={!!helpMessage}
+      aria-describedby={`${id}-error`}
       {...attributes}
     />
   );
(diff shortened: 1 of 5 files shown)


packages/form/src/Field.js
```diff
@@ -57,6 +57,7 @@ const Field = ({
       readOnly={readOnly}
       feedback
       help={!!helpMessage}
+      aria-describedby={`${id}-error`}
```
jordan-a-young: We should make this one conditional as well. You can import `useField` from `formik` to get the error state.

```
const [, { error }] = useField(id);
```

Then we can do use that error to check if we should pass in the id.

packages/form/src/Field.js
```diff
@@ -81,7 +82,7 @@ const Field = ({
   }
 
   const help = helpMessage ? <FormText id={`${id}-helpmessage`.toLowerCase()}>{helpMessage}</FormText> : null;
-  const feedback = <Feedback name={id} />;
+  const feedback = <Feedback id={`${id}-error`} name={id} />;
```
jordan-a-young: It looks like the `Feedback` component is already generating an `id` so let's use that one, and pass it in to the `aria-describedby` prop
https://github.com/Availity/availity-react/blob/master/packages/form/src/Feedback.js#L7

We will need to do this for the `SelectField` since it uses the same `Feedback` component. You can leave the `AvSelectField` as is since it uses a different one.

packages/select/src/SelectField.js
```diff
@@ -43,8 +43,8 @@ const SelectField = ({
   return (
     <FormGroup className={groupClass} for={name} disabled={attributes.disabled}>
       {thisLabel}
-      <Select name={name} feedback helpMessage={helpMessage} required={required} {...attributes} />
-      <Feedback className={classNames('d-block', feedbackClass)} name={name} />
+      <Select name={name} feedback helpMessage={helpMessage} required={required} aria-describedby={`${id}-error`} {...attributes} />
```
jordan-a-young: Same comments here as with the `Field` component. Check for the error before passing to `aria-describedby`  and the id for `Feedback`

LauRoxx:
I know this PR has collected a bit of dust, but the aria-describedby for the input and select components are already set to the error message so there's no need to add it to their corresponding field and selectfield components. 

avselect would still need these changes but the current version of react-select will not let you pass in the aria-describedby without replacing the input component which would interfere with the aria-live sections this component relies on. You can refer to [this PR](https://github.com/Availity/availity-react/pull/907) for how we handled it for the select component if you're interested in transferring it to the avselect component. idk if you'd prefer to revert your changes on this PR or close it and start fresh.

@PavanGM23 @jordan-a-young

LauRoxx:
Completed by PR #1151 
https://github.com/Availity/availity-react/pull/1151

PR #1006, opened 2022-02-17
Feat/button: Add button library with LoadingButton component

smashleez:
We have built some component in our RCM specific libraries that can be placed here, as they are not product specific and potentially can be utilized by others. 

This is a new library, eventually could contain other specialized buttons. Currently, it contains a LoadingButton which shows a loading indicator within the button onLoading. This is a common component we utilize a lot and potentially can be utilized by others so just adding it here.


packages/button/src/LoadingButton.tsx
```diff
@@ -0,0 +1,31 @@
...
+    type={type}
+    className={`button ${isLoading ? 'is-loading' : ''} ${className || ''}`}
+    {...rest}
+  >
+    <BlockUi tag="div" blocking={isLoading} keepInView />
+    {!isLoading && <>{children}</>}
+  </Button>
+);
+
+LoadingButton.defaultProps = {
```
chrishavekost: We already have a lot of components using this style of `Component.defaultProps`, but it has been deprecated and we should start moving away from it in favor of default assignments

https://github.com/facebook/react/pull/16210
https://github.com/yannickcr/eslint-plugin-react/issues/2396#issuecomment-539184761

chrishavekost (changes requested):
LGTM I just had one comment about moving to default assignments for props. Thanks for adding this to the library 😄

packages/button/src/LoadingButton.tsx
smashleez: Good to know!

chrishavekost (approved):

PR #1007, opened 2022-02-17
Feat/badge: add badge library and add RemovableBadge Component

smashleez:
We have built some component in our RCM specific libraries that can be placed here, as they are not product specific and potentially can be utilized by others. You might see more components being added here shortly :)

This is a library for badge components, which contains removable badge component (or dismissible) that utilizes the ReactStrap badge. We use these for search filters, where removing them kicks off a search.


packages/badge/.npmignore
```diff
@@ -0,0 +1,2 @@
+**/*test.js
```
jordan-a-young: Make sure to add `**/*test.tsx` as well

packages/badge/package.json
```diff
@@ -0,0 +1,38 @@
...
+  "dependencies": {
+    "@availity/icon": "^0.10.2",
+    "lodash": "^4.17.21"
+  },
+  "devDependencies": {
+    "react": "^17.0.2",
+    "react-dom": "^17.0.2"
+  },
+  "peerDependencies": {
+    "react": ">=16.3.0"
```
jordan-a-young: we should list `reactstrap` as a peer dep too

jordan-a-young (changes requested):
Overall looks good! Left two small comments. One about npmignore and the other for peer dep

packages/badge/package.json
```diff
@@ -0,0 +1,39 @@
...
+  "module": "dist/index.js",
+  "types": "dist/index.d.ts",
+  "dependencies": {
+    "@availity/icon": "^0.10.2",
+    "lodash": "^4.17.21"
+  },
+  "devDependencies": {
+    "react": "^17.0.2",
+    "react-dom": "^17.0.2",
+    "reactstrap": "^8.0.0"
```
jordan-a-young, line 31: peerDep too

jordan-a-young (approved):

PR #1009, opened 2022-02-18
feat(button): add peerdeps and change from default export

smashleez:
Adding Peer Dependencies and switching the LoadingButton from being a default export

### DIFF
--- a/packages/button/src/index.ts
+++ b/packages/button/src/index.ts
@@ -1,1 +1,1 @@
-export { default } from './LoadingButton';
+export { default as LoadingButton } from './LoadingButton';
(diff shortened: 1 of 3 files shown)


packages/button/src/index.ts
```diff
@@ -1 +1 @@
-export { default } from './LoadingButton';
\ No newline at end of file
+export { default as LoadingButton } from './LoadingButton';
```
jordan-a-young, line 2: We can turn off the eslint rule in this file. I think it will stop the build or we could ignore it for now

jordan-a-young (approved):

PR #1017, opened 2022-02-22
Bootstrapless favorites heart

danielpickett:
Test are not yet passing, so I'll keep this PR in draft until I fix them.

- Remove all Bootstrap dependencies
- Add React Query
- Refactor to TypeScript
- Prevent interaction with FavoritesHeart while mutations are pending, which fixes a bug observable in production now where clicking multiple hearts in rapid sequence on a slow connection results in unexpected behavior (changes not persisting).


danielpickett:
Oh, and this also implements CSS Modules into the repos rollup and storybook configs.

packages/favorites/src/FavoriteHeart.tsx
```diff
@@ -0,0 +1,51 @@
...
+}): JSX.Element => {
+  const { isFavorited, isDisabled, isActiveMutation, toggleFavorite } = useFavorites(id);
+
+  const handleChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+    if (onChange) onChange(isFavorited, e);
+    toggleFavorite();
+  };
+
+  return (
+    <>
```
jordan-a-young: Looks like you don't _technically_ need a fragment here. Did you have it there for some reason?

packages/favorites/src/FavoriteHeart.tsx
```diff
@@ -0,0 +1,51 @@
...
+  onMouseDown,
+}: {
+  id: string;
+  name: string;
+  onChange?: (isFavorited: boolean, event: React.ChangeEvent<HTMLInputElement>) => void;
+  onMouseDown?: (event: React.MouseEvent<HTMLInputElement, MouseEvent>) => void;
+}): JSX.Element => {
+  const { isFavorited, isDisabled, isActiveMutation, toggleFavorite } = useFavorites(id);
+
+  const handleChange = (e: React.ChangeEvent<HTMLInputElement>) => {
```
jordan-a-young: I'm personally not a fan of single letter variables. I'm good with using it if no one else has issues

packages/favorites/src/Icons.tsx
```diff
@@ -0,0 +1,37 @@
+import React from 'react';
+
+type IconProps = { className?: string; size?: string | number; color?: string };
+
+const IconBase = ({ className, size, color, pathData }: IconProps & { pathData: string }): JSX.Element => (
```
jordan-a-young: We need to make sure we cover accessibility too

packages/favorites/src/FavoritesContext.tsx
```diff
@@ -0,0 +1,128 @@
...
+  const isDisabled = status === 'loading' || status === 'idle' || activeMutationId !== null;
+
+  return {
+    isFavorited,
+    status,
+    // try to put spinner on the button that was clicked, and subtly disable the other buttons
+    isDisabled,
+    isActiveMutation,
+    toggleFavorite: isDisabled ? asyncNoOp : toggleFavorite,
+    allFavorites: favorites,
```
jordan-a-young: Why `allFavorites` and not `favorites` here?

packages/favorites/src/utils.ts
```diff
@@ -0,0 +1,87 @@
...
+  Boolean(typeof arg?.id === 'string' && typeof arg?.pos === 'number');
+
+export const validateFavorites = (unvalidatedFavorites: any) => {
+  const validatedFavorites = Array.isArray(unvalidatedFavorites) ? unvalidatedFavorites?.filter(isFavorite) : [];
+  return validatedFavorites;
+};
+
+type SettingsResponseType = { data: { favorites: Favorite[] } };
+
+type MutationVarialbes = {
```
jordan-a-young: typo

packages/favorites/src/FavoriteHeart.tsx
chrishavekost: I'm in favor of renaming to something like `event`

packages/favorites/src/utils.ts
```diff
@@ -0,0 +1,87 @@
...
+        queryClient.setQueryData('favorites', prevFavorites);
+      };
+    },
+    onSuccess(...args) {
+      const [data] = args;
+      queryClient.setQueryData('favorites', data.favorites);
+      if (options?.onSuccess) options.onSuccess(...args);
+    },
+    onError(...args) {
+      const recover = args[2];
```
chrishavekost: I think it's more clear if we keep the variables for these functions separate instead of condensing and spreading `args`, that way you don't need to look it up or debug to see what's available
`onMutate` just has a `(variables)` param
`onSuccess` has `(data, variables, context)`
`onError` has `(err, variables, context)`

https://react-query.tanstack.com/reference/useMutation

packages/favorites/src/utils.ts
```diff
@@ -0,0 +1,87 @@
...
+
+type MutationVarialbes = {
+  favorites: Favorite[];
+  activeMutationId: string;
+};
+
+const submit = ({ favorites, activeMutationId }: MutationVarialbes) =>
+  avSettingsApi
+    .setApplication(NAV_APP_ID, { favorites })
+    .then(({ data }: SettingsResponseType) => ({ favorites: data.favorites, activeMutationId }));
```
chrishavekost: Style wise I think we prefer `async`/`await` over `Promise.then()`

rollup.config.js
```diff
@@ -38,7 +38,12 @@ export default () =>
             exclude: [`${pkgPath}/src/**/*.test.tsx`, `${pkgPath}/src/**/*.test.ts`, `${pkgPath}/src/**/*.stories.tsx`],
           },
         }),
-        postcss(),
+        postcss({
+          extract: false,
+          modules: {
+            generateScopedName: '[hash:base64]',
```
chrishavekost: Should we limit the length of the hash here? Not sure what the default length would be

chrishavekost (changes requested):

packages/favorites/src/FavoritesContext.tsx
danielpickett: This was originally meant to be temporary. I wanted to render a list of all my favorites, and I realized react-query already had that data. I don't have a good reason why I called it 'allFavorites'. I guess just because the rest of the hook centered around data for the specific favorite for the id passed in. 

I'm inclined to just take this out. Do you see any value in returning all the favorites from this hook? If it does have value, I think instead we should just provide a separate `useFavoritesQuery` export, which would just be the useQuery for favorites that I already have.

What do you think?

jordan-a-young: I think that's a good idea to have a generic one that returns _all_ of them, and then a more specific hook that handles the specific logic

packages/favorites/src/Icons.tsx
danielpickett: Good call. This comment led me down a long but productive path. 

In the icons themselves, I've only added an `aria-hidden` prob, since these icons are just used for a visual restyling of a checkbox, screen readers only need to know about the checkbox, not the icons.

But the more substantial accessibility concern here was that the `loading` spinner didn't announce itself, so given the async nature of this interaction, on extra slow connections the checkbox would be inexplicably unresponsive for a couple seconds. I've made updates so that the checkbox will initially announce, "unchecked, checkbox", then after clicking it will announce "loading", then after the async all is returned it will announce, "checked, checkbox". I'm pretty happy with this on VoiceOver on Mac, but I haven't tested with other screen readers.

I'll be pushing these changes shortly.

packages/favorites/src/FavoritesContext.tsx
danielpickett: Turns out I was already exporting `useFavoritesQuery` from the package. So I'm just removing the 'allFavorites' property here. I'm not currently using that property from the useFavorites hook, and I'm not sure how much use there is for it outside of its internal use within the package.

packages/favorites/src/utils.ts
danielpickett: I like the more explicit approach you're suggesting. I ended up refactoring in a way that removed this merging of callbacks anyway. So the offending code is gone 😄

rollup.config.js
danielpickett: It ends up like this: 

image

So that's 20 characters by my count. Yeah, that does seem excessive to me. Do you have a suggestion @chrishavekost? 

I followed [this recommendation from the docs for CSS Loader](https://webpack.js.org/loaders/css-loader/#localidentname): 

> Recommendations:
> 
> - use '[path][name]__[local]' for development
> - use '[hash:base64]' for production

@jordan-a-young and I talked about this a bit. We both liked the idea of leaving the file and original class name in place, with a much smaller hash. So in this case it would be something like `FavoriteHeart__icons__BFYtN`.

I like this because it will make debugging easier for our devs. Plus it feels very familiar to me from my use of BEM. Given that these components will be in the development environments of our devs as they work, obscuring the class names with nothing but a hash seems less than ideal. Unless we publish production and development versions of our components (like React does), it will be difficult for devs to identify in DevTools which DOM elements are coming from our library components.

The only downside to this right now is that the convention for CSS Modules calls for filenames like `FavoriteHeart.module.scss`. This means that `generateScopedName: '[name]__[local]__[hash:base64:5]'` would produce a class name like `FavoriteHeart-module__icons__BFYtN`. This extra '-module' bit seems pretty ugly to me for production code. Create React App goes to some lengths to remove this in their [react-scripts webpack config](https://github.com/facebook/create-react-app/blob/5614c87bfbaae0ce52ac15aedd2cd0f91ffd420d/packages/react-scripts/config/webpack.config.js#L25) with [some fancy footwork](https://github.com/facebook/create-react-app/blob/5614c87bfbaae0ce52ac15aedd2cd0f91ffd420d/packages/react-dev-utils/getCSSModuleLocalIdent.js#L38-L39) in their react-dev-utils package.

In summary, I have given some thought to this, but the best solution will probably take some more discussion and research and is likely out of scope for my ticket. I'm definitely interested in your overall thoughts on this. In the short term, if you'd like to make a suggestion for the number of characters for the hash, I'm all ears.

packages/favorites/src/context/utils.ts
```diff
@@ -0,0 +1,79 @@
...
+};
+
+export const openMaxModal = (): void => {
+  const atMaxLog = {
+    category: 'favorites',
+    label: 'max-favorites-modal',
+    event: 'modal-open',
+  };
+
+  avLogMessagesApi.info(atMaxLog);
```
chrishavekost: What do you think about adding the user's `akaname` to the log?

packages/favorites/src/FavoriteHeart.tsx
danielpickett: Not anymore. It was there for another little testing div that has been removed. The fragment is also now removed.

packages/favorites/src/FavoriteHeart.tsx
danielpickett: Fixed

rollup.config.js
danielpickett: I changed it to `[hash:base64:12]`

chrishavekost (approved):

PR #1027, opened 2022-03-02
Refactor/spaces link top apps

gregmartDOTin:

### DIFF
--- a/packages/spaces/src/helpers.js
+++ b/packages/spaces/src/helpers.js
@@ -1,4 +1,7 @@
 import qs from 'query-string';
+import avMessage from '@availity/message-core';
+import AvLocalStorage from '@availity/localstorage-core';
+import dayjs from 'dayjs';
 
 export const INITIAL_STATE = {
   spaces: [],
@@ -70,4 +73,56 @@ export const updateUrl = (url, key, value) => {
   });
 
   return `${uri}?${newParams}`;
+};
+
+/**
+ * Top Apps
+ */
+
+const localStorageCore = new AvLocalStorage();
+
+const TOP_APPS = {
+  ALLOWED_TYPES: ['application', 'linkout', 'navigation'],
+  BLACKLIST: ['reporting', 'how_to_guide_dental_providers', 'my_account_profile', 'my_administrators'],
+  KEYS: {
+    LAST_UPDATED: 'top-apps-updated',
+    VALUES: 'myTopApps',
+  },
+  UPDATE_EVENT: 'av:topApps:updated',
+};
+
+const canTrackSpace = (spaceId, type) =>
+  TOP_APPS.ALLOWED_TYPES.some((t) => t === type) && !TOP_APPS.BLACKLIST.some((id) => id === spaceId);
+
+const getLocalStorageTopApps = (userId) => {
+  const topAppsValues = localStorageCore.get(`${TOP_APPS.KEYS.VALUES}-${userId}`);
+
+  return topAppsValues;
+};
+
+export const updateTopApps = async (spaceId, type, akaname) => {
+  if (!spaceId || !type) return;
+
+  // If we can track the space
+  if (canTrackSpace(spaceId, type)) {
+    const today = dayjs();
+
+    // Grab the current top apps from localstorage
+    const values = (await getLocalStorageTopApps(akaname)) || {};
+
+    // Update the last updated date. For use in top nav to actually sync with settings api
+    localStorageCore.set(`${TOP_APPS.KEYS.LAST_UPDATED}-${akaname}`, today.format());
+
+    const currentCount = values[spaceId] && typeof values[spaceId].count === 'number' ? values[spaceId].count : 0;
+
+    values[spaceId] = {
+      ...values[spaceId],
+      count: currentCount + 1,
+      lastUse: today.format(),
+    };
+
+    localStorageCore.set(`${TOP_APPS.KEYS.VALUES}-${akaname}`, values);
+
+    avMessage.send(TOP_APPS.UPDATE_EVENT);
+  }
 };
--- a/packages/spaces/src/helpers.js
+++ b/packages/spaces/src/helpers.js
@@ -89,14 +89,16 @@ const TOP_APPS = {
 };
 
 const getItemLocalStorage = (key) => {
-  const value = window.localStorage.getItem(key);
-  let output;
   try {
-    output = JSON.parse(value);
+    const value = window.localStorage.getItem(key);
+    if (value === null) {
+      return null;
+    }
+
+    return JSON.parse(value);
   } catch {
-    output = value;
+    return null;
   }
-  return output;
 };
 
 const canTrackSpace = (spaceId, type) =>
--- a/packages/spaces/src/helpers.js
+++ b/packages/spaces/src/helpers.js
@@ -1,6 +1,5 @@
 import qs from 'query-string';
 import avMessage from '@availity/message-core';
-import AvLocalStorage from '@availity/localstorage-core';
 import dayjs from 'dayjs';
 
 export const INITIAL_STATE = {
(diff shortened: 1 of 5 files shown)


packages/spaces/src/helpers.js
```diff
@@ -1,4 +1,7 @@
 import qs from 'query-string';
+import avMessage from '@availity/message-core';
+import AvLocalStorage from '@availity/localstorage-core';
```
chrishavekost: `@availity/localstorage-core` was [removed](https://github.com/Availity/sdk-js/commit/0dfa27385d87e356ce4e2b7a6173d2b1370eb78a) a while back in favor of just using browser local storage instead, can we refactor this to something similar?
Here's [another PR for reference](https://github.com/Availity/availity-react/commit/ed66c2397e9edd5bb43144aadf913b8c21bbb15c)

packages/spaces/src/helpers.js
```diff
@@ -71,3 +74,55 @@ export const updateUrl = (url, key, value) => {
...
+    LAST_UPDATED: 'top-apps-updated',
+    VALUES: 'myTopApps',
+  },
+  UPDATE_EVENT: 'av:topApps:updated',
+};
+
+const canTrackSpace = (spaceId, type) =>
+  TOP_APPS.ALLOWED_TYPES.some((t) => t === type) && !TOP_APPS.BLACKLIST.some((id) => id === spaceId);
+
+const getLocalStorageTopApps = (userId) => {
```
chrishavekost: I think we should rename this param to `akaname` since it's referenced that way elsewhere, and `userId` is a different value that exists

packages/spaces/src/helpers.js
```diff
@@ -71,3 +74,55 @@ export const updateUrl = (url, key, value) => {
...
+
+export const updateTopApps = async (spaceId, type, akaname) => {
+  if (!spaceId || !type) return;
+
+  // If we can track the space
+  if (canTrackSpace(spaceId, type)) {
+    const today = dayjs();
+
+    // Grab the current top apps from localstorage
+    const values = (await getLocalStorageTopApps(akaname)) || {};
```
chrishavekost: being nitpicky but I think renaming this to something like `topApps` is clearer

packages/spaces/src/helpers.js
```diff
@@ -71,3 +74,55 @@ export const updateUrl = (url, key, value) => {
...
+
+    values[spaceId] = {
+      ...values[spaceId],
+      count: currentCount + 1,
+      lastUse: today.format(),
+    };
+
+    localStorageCore.set(`${TOP_APPS.KEYS.VALUES}-${akaname}`, values);
+
+    avMessage.send(TOP_APPS.UPDATE_EVENT);
```
chrishavekost, line 136: I'm not totally sure this is necessary, let's meet tomorrow and discuss it.

chrishavekost (changes requested):

packages/spaces/src/helpers.js
gregmartDOTin: Done

packages/spaces/src/helpers.js
gregmartDOTin: done

packages/spaces/src/helpers.js
gregmartDOTin: done

packages/spaces/src/helpers.js
gregmartDOTin: For documentation purposes:

Met with Chris. We decided that for the current iteration, this is still necessary.

packages/spaces/src/helpers.js
```diff
@@ -71,3 +73,64 @@ export const updateUrl = (url, key, value) => {
...
+  BLACKLIST: ['reporting', 'how_to_guide_dental_providers', 'my_account_profile', 'my_administrators'],
+  KEYS: {
+    LAST_UPDATED: 'top-apps-updated',
+    VALUES: 'myTopApps',
+  },
+  UPDATE_EVENT: 'av:topApps:updated',
+};
+
+const getItemLocalStorage = (key) => {
+  const value = window.localStorage.getItem(key);
```
chrishavekost: `localStorage.getItem()` will return `null` if the item doesn't exist, and we should be able to return the value without creating an intermediate variable. i think we can change the function to something like this:

```js
const getItemLocalStorage = (key) => {
  try {
    const value = window.localStorage.getItem(key);
    
    // avoids JSON.parse() overhead if value already null
    if(value === null) {
      return null;
    }
    return JSON.parse(value);
  } catch {
    return null;
  }
};
```

packages/spaces/src/helpers.js
```diff
@@ -71,3 +73,64 @@ export const updateUrl = (url, key, value) => {
...
+
+export const updateTopApps = async (spaceId, type, akaname) => {
+  if (!spaceId || !type) return;
+
+  // If we can track the space
+  if (canTrackSpace(spaceId, type)) {
+    const today = dayjs();
+
+    // Grab the current top apps from localstorage
+    const topApps = (await getLocalStorageTopApps(akaname)) || {};
```
chrishavekost, line 121: Instead of defaulting `topApps` to `{}`, we can add a null check here since `getLocalStorageTopApps` and `getItemLocalStorage` can both return `null`. If there are no `topApps` from local storage to update, we can skip the rest of this logic instead of running it on an empty object

packages/spaces/src/useLink.js
```diff
@@ -26,6 +26,8 @@ export default (spaceOrSpaceId, { clientId: propsClientId, linkAttributes = {} }
     window.open(`/web/spc/disclaimers/#/?spaceId=${id}&ssoText=${metadata.disclaimerId}`, '_self');
   };
 
+  const { data: user } = useCurrentUser();
```
chrishavekost, line 29: I know me and @jordan-a-young eventually want to move as many things as we can in `availity-react` and `sdk-js` to use `react-query`, but this will be a breaking change since users will need to wrap their app in a query context provider. I think for now let's get the user info with the [sdk-js resource](https://github.com/Availity/sdk-js/blob/master/packages/api-axios/src/resources/user.js#L12) and we can move this to our react-query hooks in the future.

AFAICT the rest of the SpacesLink changes have been made backwards compatible so far, so that's why I'm hesitant to introduce react-query here. What do you all think? If we add it here, then most of the fetching logic in `SpacesLink` could/should be rewritten to take advantage of it IMO, which is probably out of scope for this PR

chrishavekost (changes requested):

packages/spaces/src/useLink.js
gregmartDOTin: Yeah, that makes sense. However, the ticket I had specifically recommended moving to userCurrentUser(). Also, I do have a ticket in the epic to convert to react-query.

packages/spaces/src/helpers.js
gregmartDOTin: That makes sense, probably ought to make that change in @availity/pagination as well: https://github.com/Availity/availity-react/blob/master/packages/pagination/src/Pagination.js#L11-L20

I can do that separately.

chrishavekost: Nice catch, yeah that's a good idea

packages/spaces/src/helpers.js
gregmartDOTin: Done

packages/spaces/src/useLink.js
gregmartDOTin: We've decided that at least for the time-being we're going to move forward with the react-query solution.

packages/spaces/src/helpers.js
chrishavekost: Ah I missed the part where we'd actually still have a current `spaceId` to update `topApps` with, so ignore me 😅

gregmartDOTin: lol, alright. I reverted that piece but kept the refactor of getItemLocalStorage

chrishavekost (approved):

PR #1035, opened 2022-03-07
refactor(pagination): make getItemLocalStorage more efficient

gregmartDOTin:
@chrishavekost noticed this while reviewing an @availity/spaces PR: https://github.com/Availity/availity-react/pull/1027#r820792870

TL;DR - `localStorage.getItem()` will return null if the item doesn't exist, and we should be able to return the value without creating an intermediate variable.

### DIFF
--- a/packages/page-header/CHANGELOG.md
+++ b/packages/page-header/CHANGELOG.md
@@ -3,6 +3,17 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+## 11.1.26 (2022-03-04)
+
+
+### Bug Fixes
+
+* **page-header:** dont send in empty payerId ([9b9e551](https://github.com/Availity/availity-react/commit/9b9e5512216cefffe2e3f67ca223e50e805dfc90))
+
+
+
+
+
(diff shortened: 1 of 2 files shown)


packages/pagination/src/Pagination.js
```diff
@@ -9,14 +9,15 @@ export const PaginationContext = React.createContext();
 export const usePagination = () => useContext(PaginationContext);
 
 const getItemLocalStorage = (key) => {
```
gregmartDOTin: Having made this change, if you look on lines [73](https://github.com/Availity/availity-react/blob/bed012eefc88f58802722be8746c82c3f2644ce2/packages/pagination/src/Pagination.js#L73) and then [96](https://github.com/Availity/availity-react/blob/bed012eefc88f58802722be8746c82c3f2644ce2/packages/pagination/src/Pagination.js#L96) the check seems fairly useless. Feels like we're setting 'current-page' to the value of `currentPage` and then checking to see if 'current-page' equals `currentPage`. Am I missing something?

jordan-a-young: I think you're right. It should be equal every time in this scenario. Might be a relic from the past? I think it would be safe to remove

PR #1036, opened 2022-03-08
refactor(pagination): remove unnecessary getItemLocalStorage check

gregmartDOTin:
continuation of https://github.com/Availity/availity-react/pull/1035, which I accidentally closed...

It turns out getItemLocalStorage is not really necessary


jordan-a-young (changes requested):
Looks good. Can you also remove where the value gets set in local storage?

gregmartDOTin:
> Looks good. Can you also remove where the value gets set in local storage?

Done

jordan-a-young (approved):

PR #1046, opened 2022-03-30
Feat/json viewer

bjnewman:
As promised, the json viewer.

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -14,3 +14,4 @@ rules:
   unicorn/prefer-array-find: 0
   unicorn/no-abusive-eslint-disable: 0
   unicorn/no-lonely-if: 0
+  unicorn/no-array-for-each: 0
--- a/packages/json-viewer/.npmignore
+++ b/packages/json-viewer/.npmignore
@@ -1,0 +1,3 @@
+**/*test.js
+**/*.stories.tsx
+**/*.stories.tsx
--- a/packages/json-viewer/.npmignore
+++ b/packages/json-viewer/.npmignore
@@ -1,3 +1,3 @@
-**/*test.js
+**/*test.tsx
 **/*.stories.tsx
 **/*.stories.tsx
--- a/packages/json-viewer/README.md
+++ b/packages/json-viewer/README.md
@@ -1,0 +1,25 @@
+# @availity/json-viewer
+
+> Simple component that renders json as an accessible list of detail tags with summaries from the keys which allows accessible exploration of raw data
+
+[![Version](https://img.shields.io/npm/v/@availity/link.svg?style=for-the-badge)](https://www.npmjs.com/package/@availity/json-viewer)
+[![NPM Downloads](https://img.shields.io/npm/dt/@availity/link.svg?style=for-the-badge)](https://www.npmjs.com/package/@availity/json-viewer)
+[![Dependency Status](https://img.shields.io/librariesio/release/npm/@availity/link?style=for-the-badge)](https://github.com/Availity/availity-react/blob/master/packages/link/package.json)
+
+## Installation
+
+### NPM
+
+```bash
+npm install @availity/json-viewer
+```
+
+### Yarn
+
+```bash
+yarn add @availity/json-viewer
+```
+
+## Documentation
+
+Check out more documentation at [availity.github.io](https://availity.github.io/availity-react/components/jsonviewer)
--- a/packages/json-viewer/README.md
+++ b/packages/json-viewer/README.md
@@ -22,4 +22,4 @@ yarn add @availity/json-viewer
 
 ## Documentation
 
-Check out more documentation at [availity.github.io](https://availity.github.io/availity-react/components/jsonviewer)
+Check out more documentation at [availity.github.io](https://availity.github.io/availity-react/components/json-viewer)
--- a/packages/json-viewer/README.md
+++ b/packages/json-viewer/README.md
@@ -2,9 +2,9 @@
 
 > Simple component that renders json as an accessible list of detail tags with summaries from the keys which allows accessible exploration of raw data
 
-[![Version](https://img.shields.io/npm/v/@availity/link.svg?style=for-the-badge)](https://www.npmjs.com/package/@availity/json-viewer)
-[![NPM Downloads](https://img.shields.io/npm/dt/@availity/link.svg?style=for-the-badge)](https://www.npmjs.com/package/@availity/json-viewer)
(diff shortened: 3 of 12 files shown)


packages/json-viewer/src/JsonViewer.tsx
```diff
@@ -0,0 +1,67 @@
...
+function getDetails({ data, expandAll, keyClassNames, summaryClassNames }: Detailable): (JSX.Element | null)[] {
+  return Object.entries(data).map(entry => {
+    const [key, value] = entry;
+    if (isPrimitive(value)) {
+      return (
+        <li key={key} className="pt-1">
+          <span className={classnames('text-bold', keyClassNames)}>
+            {`${key}:`}
+          </span>
+          <span> {(value as string | number | boolean).toString()}</span>
```
jordan-a-young: I don't think you need the `as string...` part here since you have the `isPrimitive` check

packages/json-viewer/package.json
```diff
@@ -0,0 +1,46 @@
...
+    "json"
+  ],
+  "homepage": "https://availity.github.io/availity-react/components/json-viewer",
+  "bugs": {
+    "url": "https://github.com/Availity/availity-react/issues"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/Availity/availity-react.git",
+    "directory": "packages/link"
```
jordan-a-young: some leftover copypasta

packages/json-viewer/src/JsonViewer.tsx
```diff
@@ -0,0 +1,67 @@
...
+function JsonViewer({ data, expandAll, listClassNames, keyClassNames, summaryClassNames }: JsonViewerProps): JSX.Element {
+  const details = getDetails({ data, expandAll, keyClassNames, summaryClassNames });
+  return (
+    <ul data-testid='topLevelUl' className={classnames("p-2", listClassNames)} style={{ listStyle: 'none', backgroundColor: '#fbfbfb' }}>
+      {details}
+    </ul>
+  );
+}
+
+JsonViewer.defaultProps = {
```
jordan-a-young: Instead of the `defaultProps` api, you can add this to the destructuring on line 54

packages/json-viewer/src/JsonViewer.tsx
```diff
@@ -0,0 +1,67 @@
...
+    if (isObject(value)) {
+      return (
+        <li key={key} className="pt-1">
+          <details open={expandAll}>
+            <summary className={classnames('text-bold', summaryClassNames)}>
+              {`${key}: ${
+                Array.isArray(value) ? `[ ] ${value.length} items` : `{ } ${Object.keys(value).length} keys`
+              }`}
+            </summary>
+            <ul className="pl-4" style={{ listStyle: 'none' }}>
```
jordan-a-young: instead of the style prop, let's use the className from Bootstrap. it's `list-unstyled`

packages/json-viewer/src/JsonViewer.tsx
```diff
@@ -0,0 +1,67 @@
...
+      );
+    }
+    return null;
+  });
+}
+
+function JsonViewer({ data, expandAll, listClassNames, keyClassNames, summaryClassNames }: JsonViewerProps): JSX.Element {
+  const details = getDetails({ data, expandAll, keyClassNames, summaryClassNames });
+  return (
+    <ul data-testid='topLevelUl' className={classnames("p-2", listClassNames)} style={{ listStyle: 'none', backgroundColor: '#fbfbfb' }}>
```
jordan-a-young: same thing about using `list-unstyled`. I think we should have people use the className prop to set the background color on their own.

packages/json-viewer/__tests__/JsonViewer.test.tsx
```diff
@@ -0,0 +1,43 @@
+import React from 'react';
+import { render, cleanup } from '@testing-library/react';
+import JsonViewer from '..';
+
+afterEach(cleanup);
```
jordan-a-young: testing library will run the cleanup for you so we don't need to do it manually

packages/json-viewer/README.md
```diff
@@ -0,0 +1,25 @@
+# @availity/json-viewer
+
+> Simple component that renders json as an accessible list of detail tags with summaries from the keys which allows accessible exploration of raw data
+
+[![Version](https://img.shields.io/npm/v/@availity/link.svg?style=for-the-badge)](https://www.npmjs.com/package/@availity/json-viewer)
+[![NPM Downloads](https://img.shields.io/npm/dt/@availity/link.svg?style=for-the-badge)](https://www.npmjs.com/package/@availity/json-viewer)
+[![Dependency Status](https://img.shields.io/librariesio/release/npm/@availity/link?style=for-the-badge)](https://github.com/Availity/availity-react/blob/master/packages/link/package.json)
```
jordan-a-young: last part of the url needs to go from `packages/link/package.json` -> `packages/json-viewer/package.json`

packages/json-viewer/.npmignore
```diff
@@ -0,0 +1,3 @@
+**/*test.js
```
jordan-a-young: make sure to include the `tsx` test files as well

.eslintrc.yml
```diff
@@ -14,3 +14,4 @@ rules:
   unicorn/prefer-array-find: 0
   unicorn/no-abusive-eslint-disable: 0
   unicorn/no-lonely-if: 0
+  unicorn/no-array-for-each: 0
```
jordan-a-young, line 17: Can you disable this in the specific file instead of for the whole repo?

jordan-a-young (changes requested):
Looks great overall. I left comments for some smaller things to tackle.

.eslintrc.yml
bjnewman: Thank you for catching all my dumb mistakes and the other suggestions are all good but forEach is good and these unicorn rules are overly restrictive

bjnewman: in order to be constructive and not just arguing, look at the [justifications](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/no-array-for-each.md) for this lint rule. Performance - hardly ever a concern for us, readability - i would argue for of is less readable but i think that falls squarely into dev preferences, and the ability to return early - which if it applies is a decent reason but is no reason to throw away forEach entirely

jordan-a-young (approved):

PR #1059, opened 2022-04-19
refactor/table - be able to leverage Table component without provider+ enhancements

smashleez:
This PR addresses some enhancements and bug fixes to the @availity/table component.

- Get rid of `<TableProvider/>` and instead just leverage `<Table/>` directly. This will allow easier setup for those that don't intend to have child components of Table (child controls that depend on the table instance). 

- Better documentation on how to use other components in such as pagination, inside `TableControls` for access to the `TableContext`.

- Pass in more default react-table hooks (such as `usePagination`) in order to allow for easier paging. 

- Allow custom hooks to be passed into the useTable call. 

- Move rowProps and cellProps to be functions, so that rows and cells can be styled based on the data that is contained in them

- selectable tables - created a function `getCanSelectRow` that will conditionally make a row selectable based on the data contained in the row.

- fix issue where selectable columns gets displayed (sometimes) even though selectable isn't specified on the props

- Primary Table Actions - allow to toggle visibility based on record data.

We probably should bump this to 1.0.0 with this push (does setting it as 'refactor' do that?). So that people that are leveraging the current version don't get disrupted. I did include migration steps in the documents.


packages/table/src/Controls/TableControls.tsx
```diff
@@ -1,20 +1,33 @@
...
   disabled?: boolean;
   children?: React.ReactNode[] | React.ReactNode;
 } & React.HTMLAttributes<HTMLElement>;
 
-const TableControls = ({ id, disabled, children, ...rest }: Props): JSX.Element => {
+const TableControls = ({ id, disabled, children, ...rest }: Props): JSX.Element | null => {
   const [isDisabled, setIsDisabled] = useState(disabled || false);
 
+  const { instance } = useTableContext();
+  const [isReady, setIsReady] = useState(!!instance);
```
jordan-a-young: I don't think you need to use state for this. Instead, you can change your check to see if `instance` exists.

packages/table/src/Table.tsx
```diff
@@ -1,144 +1,195 @@
...
-    getTableProps,
-    getTableBodyProps,
-    headerGroups,
-    rows,
-    prepareRow,
-    selectedFlatRows: selectedRows,
-    toggleHideColumn,
-  } = instance as TableInstance<T>;
+}: TableProps<T>): JSX.Element | null => {
+  let selectionColumn: Column<T>;
```
jordan-a-young: Why not create the variable inline on 110? Was there some issue with defining it inside the function?

packages/table/src/TableCell.tsx
```diff
@@ -1,30 +1,38 @@
...
+  scrollable,
+  children,
+  onCellClick,
+  getCellProps,
+  ...rest
+}: Props<T>): JSX.Element => {
   const { className, disableClick, stickyLeft, stickyRight } = cell.column;
 
   const isFixedWidth = scrollable && !className;
+  const cellClasses = classNames(className || '', {
```
jordan-a-young: I think we can move this to line 50 where all of the classes are reconciled. That way this calculation will only have to happen when that function is called

packages/table/src/TableContent.tsx
```diff
@@ -0,0 +1,116 @@
...
+    additionalContentProps,
+
+    onRowClick,
+    onCellClick,
+    getCellProps,
+    getRowProps,
+  } = useTableContext();
+
+  const { getTableProps, getTableBodyProps, headerGroups, rows, prepareRow, manualSortBy, page } =
+    tableInstance as TableInstance<T>;
```
jordan-a-young, line 34: I've seen this pattern a few times. Is it possible to pass the generic `T` to useTableContext in a way that all of the returned variables will have the correct type?

smashleez: Believe me, I have tried....many times. I have researched how to get Context/Provider to be generic and it is a mess. Because TableContext is static we have to know what type it is before it gets created.
The closest thing that I have found to working is this: https://hipsterbrown.com/musings/musing/react-context-with-generics/ - but it assumes that when you create the Provider it knows it's concrete type (not the generic type). 

This is the cleanest thing that I have found that works.

packages/table/src/Table.tsx
smashleez: Just a bit left from the refactoring. this was used outside of the hook before.

LauRoxx:
> We probably should bump this to 1.0.0 with this push (does setting it as 'refactor' do that?). So that people that are leveraging the current version don't get disrupted. I did include migration steps in the documents.

I believe you need the ! or the BREAKING CHANGE: footer in the commit as listed in the [conventional commit docs](https://www.conventionalcommits.org/en/v1.0.0/)

smashleez:
> > We probably should bump this to 1.0.0 with this push (does setting it as 'refactor' do that?). So that people that are leveraging the current version don't get disrupted. I did include migration steps in the documents.
> 
> I believe you need the ! or the BREAKING CHANGE: footer in the commit as listed in the [conventional commit docs](https://www.conventionalcommits.org/en/v1.0.0/)

Thanks! Good old git rebase for the win....

docusaurus/docs/components/table/scrollableContainer.md
```diff
@@ -18,20 +18,16 @@ const columns = [
   {
     Header: 'Column 2',
     accessor: 'column2',
-  },
+  },s
```
jordan-a-young: typo?

jordan-a-young (approved):

PR #1071, opened 2022-05-09
Feat/cc/cc 149

celigio:

### DIFF
--- a/docusaurus/docs/form/select/components/resource-select.md
+++ b/docusaurus/docs/form/select/components/resource-select.md
@@ -269,7 +269,7 @@ const Example = () => (
       customerId={customerId}
       parameters={{
         region: 'FL',
-        tranTypeCode='1',
+        tranTypeCode: '1',
       }}
       label="Select a Payer"
       required
--- a/docusaurus/docs/form/select/components/resource-select.md
+++ b/docusaurus/docs/form/select/components/resource-select.md
@@ -184,6 +184,7 @@ The following components can be imported by name from `@availity/select/resource
 - AvNavigationSelect
 - AvUserSelect
 - AvCodeSelect
+- AvPayerSelect
 
 These components are `ResourceSelect` with pre-configured `resource`, `valueKey`, and `labelKey` to make it easy to use. All of the props for `ResourceSelect` can be provided to override the defaults of these pre-made components. For some of these components, you will want to provide the `customerId` prop.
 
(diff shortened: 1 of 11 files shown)


docusaurus/docs/form/select/components/resource-select.md
```diff
@@ -258,6 +262,18 @@ const Example = () => (
...
     <AvUserSelect name="user" label="Select a User" customerId={customerId} />
     <AvCodeSelect name="code" label="Select a Code" />
+    <AvPayerSelect
+      name="payer"
+      requiredParams={['region', 'tranTypeCode']}
+      watchParams={['region', 'tranTypeCode']}
+      customerId={customerId}
+      parameters={{
+        region: 'FL',
+        tranTypeCode='1',
```
jordan-a-young: The `=` should be a `:` here

packages/mock/src/data/extended-payers.json
```diff
@@ -0,0 +1,108 @@
...
+    "links": {
+        "self": {
+            "href": "https://qa-apps.availity.com/api/internal/v1/extended-payers?state=FL&transactionType=EBREACT"
+        }
+    },
+    "extendedAccessPayers": [
+        {
+            "id": "17178",
+            "payerName": "ALLEGIANCE BENEFIT PLAN",
+            "payerId": "81040"
```
jordan-a-young: Can you change these to be mock data like `Test Payer` and `payerId: "123"`?

packages/select/custom-resources/AvPayerSelect.js
```diff
@@ -0,0 +1,34 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import AvApi from '@availity/api-axios';
+
+import { ResourceSelect } from '..';
+
+const extendedPayersApi = new AvApi({ path: '/api/internal', name: 'extended-payers'});
+
+extendedPayersApi.all = parameters => { extendedPayersApi.query({params: { state: parameters.region, transactionType: parameters.tranTypeCode }, headers: {'X-Availity-Customer-ID': parameters.customerId}})};
```
jordan-a-young, line 9: Two things about this function. Is it used? since `pageAll` is false, I think this won't be called. Also, looks like this function doesn't return anything if it is being called

celigio: When it was set to pageAll: true it was calling the endpoint constantly and not displaying any of the returning payers. When I set it to false it would call the endpoint once and display the results.

docusaurus/docs/form/select/components/resource-select.md
celigio: yup, fixing it now

packages/mock/src/data/extended-payers.json
celigio: Done

PR #1091, opened 2022-06-01
fix(spaces): add ability to give class to Link title PF-2340

owolfhu1:
adds ability to give class to Link title

### DIFF
--- a/packages/spaces/src/SpacesLink.js
+++ b/packages/spaces/src/SpacesLink.js
@@ -227,12 +227,14 @@ const Link = ({
               <Media body id={`${type}-${configurationId}`} className="text-dark">
                 <TitleTag
                   id={`app-title-${configurationId}`}
-                  className={classNames({
-                    'mb-0': !showDescription || !description,
-                    'pt-3': stacked,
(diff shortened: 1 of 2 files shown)


packages/spaces/src/SpacesLink.js
```diff
@@ -230,6 +231,7 @@ const Link = ({
                     'mb-0': !showDescription || !description,
                     'pt-3': stacked,
                     'text-center': stacked,
+                    [titleClassName]: !!titleClassName,
```
jordan-a-young: `classnames` accepts multiple args so i think we should use that here. It will remove any falsy values
```
classNames(<existing object>, titleClassName)
```

jordan-a-young (changes requested):
Can you add a unit test for this prop please?

JoeSpanbauer (dismissed):
Not sure if the array adds any value, but it also shouldn't break it. :)

owolfhu1:
@jordan-a-young thanks for that, updated to use second classNames() param and added unit tests.

jordan-a-young (approved):

PR #1092, opened 2022-06-01
feat/tree - Add tree component to display hierarchical data

smashleez:


packages/tree/package.json
```diff
@@ -0,0 +1,52 @@
+{
+  "name": "@availity/tree",
+  "version": "1.0.0",
+  "description": "TThis is a component for displaying hierarchical data",
```
chrishavekost, line 4: Extra 'T' here

packages/tree/package.json
```diff
@@ -0,0 +1,52 @@
...
+  },
+  "devDependencies": {
+    "react": "^17.0.2",
+    "react-dom": "^17.0.2",
+    "reactstrap": "^8.10.1",
+    "tsup": "^5.12.7",
+    "typescript": "^4.6.4"
+  },
+  "peerDependencies": {
+    "react": ">=16.3.0",
```
chrishavekost, line 46: May want to specify `^16.3.0 || ^17.0.0` since react-block-ui isn't compatible with react 18. Depending on when this gets merged, the updated block ui may be available and we could use that instead

packages/tree/src/Tree.tsx
```diff
@@ -0,0 +1,381 @@
...
+
+  const areAllExpanded = (items: TreeItem[]) => {
+    let allExpanded = items.every((item) => item.isExpanded);
+    for (const item of items) {
+      if (item.children) {
+        allExpanded = allExpanded && areAllExpanded(item.children);
+      }
+    }
+    return allExpanded;
+  };
```
chrishavekost, line 72: For large trees would it make sense to memoize the result?

packages/tree/src/Tree.tsx
```diff
@@ -0,0 +1,381 @@
...
+      return item;
+    }
+
+    for (const item of items) {
+      if (item.children) {
+        return findItem(item.children, id);
+      }
+    }
+    return null;
+  };
```
chrishavekost, line 156: I'm wondering if it makes sense to move from an array of Tree items to a Map or Set of them. You can still iterate over the items when you need to, but these recursive functions where you're iterating over all the items could be replaced with an O(1) operation

Do you think it's worth it to change?

packages/tree/src/Tree.tsx
```diff
@@ -0,0 +1,381 @@
...
+                          checked={item.isSelected}
+                          onChange={() => toggleSelect(item)}
+                        />
+                      )}
+                      <Label data-testid={`tree-view-item-${item.id}-label`} className={item.isDisabled ? 'text-muted' : ''} check>
+                        {item.name}
+                      </Label>
+                    </FormGroup>
+
+                    {item.children && item.children.length > 0 && (
```
chrishavekost, line 331: This can just be `item.children?. length > 0 &&`

smashleez:
@chrishavekost this is just a draft and not ready for review yet. I thought it wasn’t supposed to notify until I opened it for review. Closing this until I am ready for it to be revised.

packages/tree/src/Tree.tsx
smashleez: I agree, but this doesn't compile for some reason. 
image

PR #1093, opened 2022-06-02
Fix/table checkboxes -  Fix issue where checkbox checks are not showing up in UI Kit v4

smashleez:
This is a workaround to an issue where the white check in the checkbox does not show up when it is selected.

The checkbox utilizes classes from availity-ukit ^4.0.0, which seems to be the issue. Version 4 loads the checks via an SVG, which is different than how it was loaded in version 3. 
http://availity.github.io/availity-uikit/v4/components#Forms-Checkboxes 

This change just updates it to not utilize those classes until we have appropriate time to research the issue. 

Issue: 
image

After Fix:
image

### DIFF
--- a/packages/table/src/IndeterminateCheckbox.tsx
+++ b/packages/table/src/IndeterminateCheckbox.tsx
@@ -37,10 +37,7 @@ const IndeterminateCheckbox = React.forwardRef(
 
     return (
       <>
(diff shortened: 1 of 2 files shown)


jordan-a-young (approved):
Looks good to me. Feel free to merge once the checks pass

chrishavekost (approved):

PR #1096, opened 2022-06-03
feat(hooks): add new hook for 508 compliant event handlers for tabs

bjnewman:
reopened this after previous PR https://github.com/Availity/availity-react/pull/1094 closed to rebase and squash multiple commits that were growing unwieldy. 
From original: Based on https://www.w3.org/WAI/ARIA/apg/patterns/tabpanel/#:~:text=.-,Keyboard%20Interaction,-For%20the%20tab we have several internal components that need to add 508 compliant keyboard and focus handlers for tab lists, commonly found using Reactstrap's nav tab components. Because of the multiple uses of these components it is difficult to build a generalized solution in component form that meets all user needs. But we will likely be able to build a limited use compatible wrapper tab component utilizing this hook.

Another design goal for this hook was to make it compatible with multiple horizontal tab list components, and multiple state management solutions.

Also certainly open to feedback on whether you think this is even the right approach.

Note that as opposed to the original PR - this one relies more on users to manage their own tabIndices to control the focus order but in return greatly simplifies the logic and eliminates multiple difficult edge cases from manually managing focus.

### DIFF
--- a/packages/hooks/src/useTabsEventHandlers.js
+++ b/packages/hooks/src/useTabsEventHandlers.js
@@ -1,0 +1,54 @@
+import * as React from 'react';
+
+function mathMod(a, b) {
+  return ((a % b) + b) % b;
+}
+function isObject(value) {
+  return !!value && typeof value === 'object';
+}
+
+export default function useTabsEventHandlers(tab, tabs, updaterFn, active, options = {}) {
+  const { customFindFn } = options;
+  if ((typeof tab !== 'string' && !isObject(tab)) || (isObject(tab) && !('name' in tab))) {
+    throw new Error('useTabsEventHandler requires tabs to be strings or objects with a name property');
+  }
+
+  if ((isObject(tab) || isObject(active)) && typeof customFindFn !== 'function') {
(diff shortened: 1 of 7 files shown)


packages/hooks/src/useTabsEventHandlers.js
```diff
@@ -0,0 +1,54 @@
...
+}
+
+export default function useTabsEventHandlers(tab, tabs, updaterFn, active, options = {}) {
+  const { customFindFn } = options;
+  if ((typeof tab !== 'string' && !isObject(tab)) || (isObject(tab) && !('name' in tab))) {
+    throw new Error('useTabsEventHandler requires tabs to be strings or objects with a name property');
+  }
+
+  if ((isObject(tab) || isObject(active)) && typeof customFindFn !== 'function') {
+    throw new Error('when using tab objects a custom find function is required');
```
bjnewman, line 17: Similar to previous pr i am interested in feedback on this strategy of throwing errors as opposed to warning and using a sensible default like if(isObject(tab) && typeof customFindFn === 'undefined' {customFindFn = (tabs, active) => tabs.find(tab => tab.name === active.name}

PR #1100, opened 2022-06-10
Add Tree Component

smashleez:
This Pull Request is to add a new package (@availity/tree) that will display a selectable, expandable/collapsible tree component for displaying hierarchical data.

image

### DIFF
--- a/packages/tree/src/Tree.tsx
+++ b/packages/tree/src/Tree.tsx
@@ -27,19 +27,21 @@ export type TreeProps = {
 
 export const buildTree = (items: TreeItem[]) => {
   const tree: TreeItem[] = [];
-  const hash: Record<string, TreeItem> = {};
+  const parents: Map<string, TreeItem> = new Map<string, TreeItem>();
 
   for (const item of items) {
-    hash[item.id] = item;
-    hash[item.id].children = [];
+    item.children =[];
+    parents.set(item.id, item);
   }
 
-  for (const id in hash) {
-    const item = hash[id];
-    if (item.parentId) {
-      hash[item.parentId]?.children?.push(item);
+  for (let [key, value] of parents) {
+    if (value.parentId && parents.get(value.parentId)) {
+      const parent = parents.get(value.parentId);
+      if (parent) {
+        parent.children?.push(value);
+      }
     } else {
-      tree.push(item);
+      tree.push(value);
     }
   }
   return tree;
--- a/packages/tree/src/Tree.tsx
+++ b/packages/tree/src/Tree.tsx
@@ -1,0 +1,381 @@
+import Icon from '@availity/icon';
+import React, { useState, useEffect, useCallback } from 'react';
+import { Button, FormGroup, Input, Label } from 'reactstrap';
+import TreeItem from './TreeItem';
+
+const SELECT_ALL = 'Select All';
+const DESELECT_ALL = 'Deselect All';
+
+const EXPAND_ALL = 'Expand All';
+const COLLAPSE_ALL = 'Collapse All';
+
+const areAllChildrenSelected = (item: TreeItem) =>
+  (item.isSelected && item.children?.every((child) => child.isSelected)) || false;
+
+export type TreeProps = {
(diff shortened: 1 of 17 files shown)


smashleez:
Not Ready for review - can't get Code Coverage generated so creating this PR to determine that.

smashleez:
Ready for review now, thanks!

packages/tree/src/Tree.tsx
```diff
@@ -0,0 +1,373 @@
...
+        walkTree(item, (item) => {
+          item.areAllChildrenSelected = areAllChildrenSelected(item);
+        });
+      }
+    }
+  }, [selectedItems, walkTree, treeItems]);
+
+  const filterItems = (items: TreeItem[], searchTerm: string) => {
+    const searchValue = searchTerm.toUpperCase();
+    const rootItems = items.filter((item) => item.name.toUpperCase().includes(searchValue));
```
jordan-a-young, line 148: I'm slightly worried about performance for larger arrays with all of the array traversal. Not a show stopper, but I wanted to bring it up sooner than later.

packages/tree/src/Tree.tsx
```diff
@@ -0,0 +1,373 @@
...
+    const itemToUpdate = treeItems.find((t) => t.id === item.id);
+    if (itemToUpdate) {
+      item.isSelected = !item.isSelected;
+      if (item.isSelected) {
+        setSelectedList([...selectedList, item]);
+      } else {
+        setSelectedList(selectedList.filter((item) => item.id !== itemToUpdate.id));
+      }
+    }
+    setTreeItems([...treeItems]);
```
jordan-a-young: What is this line used for?  I don't see where `treeItems` changed

packages/tree/src/Tree.tsx
```diff
@@ -0,0 +1,373 @@
...
+                          <Icon size="lg" className="expand-tree" name={item.isExpanded ? 'down-dir' : 'right-dir'} />
+                        </Button>
+                      </div>
+                    )}
+                  </div>
+                </>
+              )}
+              {item.children && item.children.length > 0 && item.isExpanded && (
+                <ul>
+                  <Tree
```
jordan-a-young, line 352: I like the recursive approach!

packages/tree/src/TreeItem.ts
```diff
@@ -0,0 +1,31 @@
+export default class TreeItem {
```
jordan-a-young, line 1: Why did you opt for a class instead of an object? I'm not saying you should change it. Just trying to learn.

packages/tree/src/Tree.tsx
smashleez: We have utilized this logic throughout RCM for a few years now and there hasn't been anything crazy that we have found. Obviously this might be different in the react world, and we don't have anything that super large in depth. We use this for our office hierarchies. We can definitively look into improving it in the future

packages/tree/src/Tree.tsx
smashleez: It doesn't rerender without this, pretty sure that is why I had to do this.

packages/tree/src/TreeItem.ts
smashleez: I pulled a lot of this from our angular 9 version and it was a class in there, no reason other than that. Typically in our angular Apps we utilized classes, but I am not sure there is a rhyme or reason why. Just a choice

packages/tree/src/Tree.tsx
jordan-a-young: If y'all haven't seen any problems with it then that's good enough for me.

packages/tree/src/Tree.tsx
smashleez: I reran this today removing the setTreeItems and it seems to show the tree items as selected just fine. When I built this originally a few months ago I do remember having some issues, but I don't see it now. Removed.

jordan-a-young (approved):

PR #1113, opened 2022-06-30
feat(select): move clear field button, allow aria-required, export styling fix #933

### LINKED ISSUE #933: Extract visual aspect of Select/SelectField for usage without Form/Formik
Can we extract the presentation/design of the Select/SelectField components for consistent visual usage in other aspects of an application that don't require Formik? It seems that currently, these components will not function without being wrapped in a Formik context. One common use case would be for filter dropdowns.

LauRoxx:
- removed built-in clear field link/indicator, added clear field button outside of field
- pulled style and theme objects out of component and added as single export (issue #933 )
- required prop puts aria-required on input

### DIFF
--- a/packages/select/src/Select.d.ts
+++ b/packages/select/src/Select.d.ts
@@ -1,13 +1,26 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
 import type { ReactNode } from 'react';
-import type { Props as RSelectProps, GroupBase } from 'react-select';
+import type { Props as RSelectProps, GroupBase, StylesConfig, ThemeConfig } from 'react-select';
 import type { AsyncPaginateProps } from 'react-select-async-paginate';
 import type { FieldValidator } from 'formik';
+import React from 'react';
+
+export interface SelectStyleProps {
+  showError?: boolean;
+  styles?: React.CSSProperties;
+  isInline?: boolean;
+}
+
+export declare const SelectStyles: (props: SelectStyleProps) => {
+  styles: StylesConfig;
+  theme: ThemeConfig;
+};
 
 export type SelectProps<Option, IsMulti extends boolean, Group extends GroupBase<Option> = GroupBase<Option>> = {
   allowSelectAll?: boolean;
   autofill?: boolean | Record<string, string | ((value: any) => any)>;
   cacheUniq?: any | any[];
+  clearButtonClassName?: string = 'btn btn-link link';
   creatable?: boolean;
   feedback?: boolean;
   helpMessage?: ReactNode;
@@ -18,6 +31,7 @@ export type SelectProps<Option, IsMulti extends boolean, Group extends GroupBase
   name: string;
   onChange?: (value: Option) => void;
   raw?: boolean;
+  required?: boolean;
   selectRef?: AsyncPaginateProps<Option, Group, any, IsMulti>['selectRef'];
   validate?: FieldValidator;
   waitUntilFocused?: boolean;
--- a/packages/select/src/Select.d.ts
+++ b/packages/select/src/Select.d.ts
@@ -3,15 +3,14 @@ import type { ReactNode } from 'react';
 import type { Props as RSelectProps, GroupBase, StylesConfig, ThemeConfig } from 'react-select';
 import type { AsyncPaginateProps } from 'react-select-async-paginate';
 import type { FieldValidator } from 'formik';
-import React from 'react';
 
-export interface SelectStyleProps {
+export interface SelectStyleArgs {
   showError?: boolean;
-  styles?: React.CSSProperties;
+  styles?: StylesConfig;
   isInline?: boolean;
 }
 
-export declare const SelectStyles: (props: SelectStyleProps) => {
+export declare const selectStyles: (props: SelectStyleArgs) => {
   styles: StylesConfig;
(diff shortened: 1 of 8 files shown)


packages/select/src/Select.js
```diff
@@ -75,6 +67,107 @@ const validateSelectAllOptions = (options) => {
   }
 };
 
+// needed for inline styling of clear button
+const wrapperStyle = { display: 'flex' };
+
+export const SelectStyles = (showError, styles, isInline) => ({
```
jordan-a-young: I think we should have this be `selectStyles` since uppercase normally is used for components

packages/select/src/Select.js
```diff
@@ -306,130 +403,57 @@ const Select = ({
...
+        name={name}
+        classNamePrefix="av"
+        role="listbox"
+        onCreateOption={handleCreate}
+        className={classNames(
+          className,
+          'av-select',
+          touched ? 'is-touched' : 'is-untouched',
+          fieldError ? 'av-invalid' : 'av-valid',
+          touched && fieldError && 'is-invalid'
```
jordan-a-young: let's use your new variable, `errorShown`, here as well

packages/select/src/Select.js
```diff
@@ -449,19 +473,27 @@ Select.propTypes = {
...
-  innerProps: PropTypes.object,
+components.Input.propTypes = {
+  selectProps,
 };
 
 components.MultiValueRemove.propTypes = {
   innerProps: PropTypes.object,
 };
 
+SelectStyles.propTypes = {
```
jordan-a-young: If we make `SelectStyles` lowercase, then I don't think we need to specify propTypes here or in the `.d.ts` file

packages/select/stories/select.stories.tsx
```diff
@@ -40,28 +52,48 @@ export const Default: Story = ({ autofill, creatable, disabled, helpMessage, isM
       autoFill1: '',
       autoFill2: '',
     }}
-    validationSchema={isMulti ? multiValueSchema('select', required, min, max) : singleValueSchema('select', required)}
+    validationSchema={
+      isMulti
+        ? multiValueSchema('select', required, min, max, !autofill)
```
jordan-a-young: Does `autofill` determine if we should use `string` vs `object`? I feel like `raw` should tell us this instead

packages/select/src/Select.d.ts
```diff
@@ -1,13 +1,26 @@
...
+export declare const SelectStyles: (props: SelectStyleProps) => {
+  styles: StylesConfig;
+  theme: ThemeConfig;
+};
 
 export type SelectProps<Option, IsMulti extends boolean, Group extends GroupBase<Option> = GroupBase<Option>> = {
   allowSelectAll?: boolean;
   autofill?: boolean | Record<string, string | ((value: any) => any)>;
   cacheUniq?: any | any[];
+  clearButtonClassName?: string = 'btn btn-link link';
```
jordan-a-young: I think it's ok to leave off the default value here

packages/select/src/Select.d.ts
```diff
@@ -1,13 +1,26 @@
...
 import type { ReactNode } from 'react';
-import type { Props as RSelectProps, GroupBase } from 'react-select';
+import type { Props as RSelectProps, GroupBase, StylesConfig, ThemeConfig } from 'react-select';
 import type { AsyncPaginateProps } from 'react-select-async-paginate';
 import type { FieldValidator } from 'formik';
+import React from 'react';
+
+export interface SelectStyleProps {
+  showError?: boolean;
+  styles?: React.CSSProperties;
```
jordan-a-young: Can you add `CSSProperties` to the import on line 2 instead?

jordan-a-young (changes requested):

jordan-a-young:
Functionality and code looks good overall. Left a few comments.

packages/select/stories/select.stories.tsx
LauRoxx: these are just going off the storybook args so it looks like I'll need both.

packages/select/src/Select.js
LauRoxx: oh, I don't think I've seen that. why would that be?

packages/select/src/Select.d.ts
LauRoxx: Do you mean that it's ok to leave it off because it's in the Select.defaultProps or to not have a default at all? I felt a bit weird adding bootstrap to the component when we're taking it out of so many others so that's why I just made it the default.

packages/select/src/Select.js
jordan-a-young: The `prop-types` package is for components. When `SelectStyles` is uppercase the linter probably thinks it is a component instead of a regular function. If we make it lowercase then the linter won't throw an error. We should still have the types in `.d.ts`. I jumped the gun on that one

packages/select/src/Select.d.ts
jordan-a-young: The first one. The actual value that the component is using is coming from `defaultProps`.  When you set a default in the type def, it's saying the default type is `btn btn-link link`, not the value.

jordan-a-young (dismissed):

jordan-a-young (approved):

PR #1138, opened 2022-08-10
Add support for spaces query status parameter  

gogo-jscott:

### DIFF
--- a/packages/page-header/src/PageHeader.js
+++ b/packages/page-header/src/PageHeader.js
@@ -32,6 +32,7 @@ const PageHeader = ({
   homeUrl,
   linkTag,
   className,
+  variables,
   ...props
 }) => {
   const spaces = useSpaces(spaceId, payerId);
@@ -67,6 +68,7 @@ const PageHeader = ({
           spaceIds={spaceId ? [spaceId] : undefined}
           payerIds={payerId ? [payerId] : undefined}
           clientId={clientId}
+          variables={variables}
         >
           <SpacesLogo {...logoAttrs} />
         </Spaces>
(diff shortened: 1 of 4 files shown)


packages/page-header/src/PageHeader.js
```diff
@@ -165,6 +167,7 @@ PageHeader.propTypes = {
   clientId: PropTypes.string,
   iconSrc: PropTypes.string,
   iconAlt: PropTypes.string,
+  variables: PropTypes.object,
```
jordan-a-young: Can you add this to the PageHeaderProps type as well please? The type should be `Record<string, unknown>`

jordan-a-young (changes requested):
I think it looks good other than the missing type def.

@chrishavekost @owolfhu1 @gregmartDOTin Will updating the default query have any negative impact on the spaces api?

packages/page-header/src/PageHeader.js
jordan-a-young: Also, what do you think about naming the prop `spacesVariables`? That way it will be less generic

chrishavekost:
I think @owolfhu1 is looking into how this will affect the API but I'm wondering if @gogo-jscott can explain the reasoning behind these changes and if they're needed for everyone. If it's a one-off change I think each component will let you override the default query

dnoler:
@chrishavekost I can provide context for this change. We source payer spaces for payer logos, but we have an issue on Claims where a particular payer has an inactive payer space, so their logo isn't found. Granted, we probably should have a better solution for how we source payer logos, but this gets us to where we can at least define inactive spaces for payers that don't have a proper payer space section on the portal.

owolfhu1:
Can confirm updated query is still working

gogo-jscott:
Closing since feature goes outside intended purpose of package.

PR #1140, opened 2022-08-18
Fix aria describedby daterangefield

Valar05:
Changes to DateRangeField such that it's feedback component is no longer referenced by aria-describedby unless the feedback component is visible.

### DIFF
--- a/packages/date/src/DateRange.js
+++ b/packages/date/src/DateRange.js
@@ -59,7 +59,7 @@ const DateRange = ({
   innerRef,
   className,
   format = isoDateFormat,
-  ariaDescribedBy,
+  feedback,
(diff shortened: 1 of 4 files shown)


packages/date/src/DateRange.js
```diff
@@ -323,7 +326,7 @@ DateRange.propTypes = {
   validate: PropTypes.func,
   min: limitPropType,
   max: limitPropType,
-  ariaDescribedBy: PropTypes.string,
+  feedback: PropTypes.bool,
```
jordan-a-young, line 329: Can you add this to the type definition for the props as well?

jordan-a-young (approved):

PR #1141, opened 2022-08-18
fix(pagination): update div to li to fix 508 compliance issues

smashleez:
We are focusing on fixing 508 compliance issues, and ran into the following violation with the pagination component, specifically involving the pagination text.

 **This UL should only contain li elements (without an ARIA-assigned role), script elements (without an ARIA-assigned role) or template elements (without an ARIA-assigned role); or elements with a role=listitem attribute; as direct child elements**
 
 This is because the text was in a `<div>` rather than an `<li>`.

### DIFF
--- a/packages/pagination/package.json
+++ b/packages/pagination/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@availity/pagination",
-  "version": "2.15.2",
+  "version": "2.15.3-alpha.0",
   "description": "component for displaying paginated data",
   "keywords": [
     "availity",
--- a/packages/pagination/src/PaginationControls.tsx
+++ b/packages/pagination/src/PaginationControls.tsx
@@ -153,9 +153,9 @@ const PaginationControls = ({
         ''
       )}
       {showPaginationText && (
-        <div data-testid="pagination-text" className="pagination-text pt-1 pl-2 pr-2">
+        <li data-testid="pagination-text" className="pagination-text pt-1 pl-2 pr-2">
           {populatePaginationText ? populatePaginationText(lower, upper, total) : `${lower}-${upper} of ${total}`}
-        </div>
+        </li>
       )}
     </Pagination>
   ) : null;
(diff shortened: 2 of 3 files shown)


packages/pagination/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@availity/pagination",
-  "version": "2.15.2",
+  "version": "2.15.3-alpha.0",
```
smashleez, line 3: QUESTION: When we create alpha versions it automatically creates a commit this updated alpha version. Is it okay to leave this in there or does it need to be undone?

jordan-a-young: I think it should be ok. The commit did have `[skip ci]` in the message so the PR checks are not running. We need to tweak the setup so that doesn't happen. Would you be able to edit the commit message to remove the `[skip ci]`? The checks do a dry-run of the version bump so we can confirm it will be bumped properly

packages/pagination/src/PaginationControls.tsx
```diff
@@ -153,9 +153,9 @@ const PaginationControls = ({
         ''
       )}
       {showPaginationText && (
-        <div data-testid="pagination-text" className="pagination-text pt-1 pl-2 pr-2">
+        <li data-testid="pagination-text" className="pagination-text pt-1 pl-2 pr-2">
```
jordan-a-young, line 156: I'm wondering if we should move this outside the `Pagination` component and wrap them both with a fragment. I'm not sure if it will mess with the styles/layout though. I'll have to check

smashleez: I can play with it really quickly.

smashleez: Yeah it throws the styling off, which is not impossible to fix. But it seems a bit cleaner to keep this under the `<ul>` as a `<li>` if you are cool with it.

jordan-a-young: I am cool with it assuming it's compliant

jordan-a-young (approved):

PR #1149, opened 2022-09-19
feat(AvSelectField):Add feedback message Id to be referenced by aria-…

mdiallo21:
…describedby

### DIFF
--- a/packages/reactstrap-validation-select/src/AvSelect.js
+++ b/packages/reactstrap-validation-select/src/AvSelect.js
@@ -234,6 +234,8 @@ class AvSelect extends AvBaseInput {
       creatable,
       options,
       components: componentsOverride,
+      placeholder,
+      ariaMessage,
       ...attributes
     } = this.props;
     const { newOptions } = this.state;
@@ -258,6 +260,16 @@ class AvSelect extends AvBaseInput {
     if (!attributes.loadOptions && creatable) {
       Tag = Creatable;
     }
+
+     // Enhance placeholder for accessibility
+  placeholder = (
+    <>
+      {placeholder || 'Select...'}
+      <span className="sr-only">
+        {ariaMessage.ariaErrorDesc || null} {helpMessage || null}
+      </span>
+    </>
+  );
 
     return (
       <Tag
@@ -360,6 +372,9 @@ class AvSelect extends AvBaseInput {
             primary: '#3262af',
           },
         })}
+        aria-errormessage={ariaMessage.hasError ? ariaMessage.ariaErrorId : ''}
+        aria-invalid={ariaMessage.hasError}
+        placeholder={placeholder}
         options={!attributes.loadOptions ? [...options, ...newOptions] : undefined}
         onCreateOption={this.handleCreate}
         components={{ ...components, ...componentsOverride }}
@@ -381,6 +396,8 @@ AvSelect.propTypes = {
   loadOptions: PropTypes.func,
   raw: PropTypes.bool,
   creatable: PropTypes.bool,
+  ariaMessage: PropTypes.object,
+  placeholder: PropTypes.string,
   autofill: PropTypes.oneOfType([PropTypes.bool, PropTypes.object]),
 };
 
--- a/packages/reactstrap-validation-select/src/AvSelectField.js
+++ b/packages/reactstrap-validation-select/src/AvSelectField.js
@@ -43,7 +43,7 @@ class AvSelectField extends Component {
     let feedback = null;
     if (validation.errorMessage) {
       const feedbackClasses = classNames('d-block', feedbackClass);
-      feedback = <AvFeedback className={feedbackClasses}>{validation.errorMessage}</AvFeedback>;
+      feedback = <AvFeedback className={feedbackClasses} id={`${id}-error`}>{validation.errorMessage}</AvFeedback>;
     }
 
     let thisLabel = false;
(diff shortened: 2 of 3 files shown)


packages/reactstrap-validation-select/src/AvSelectField.js
```diff
@@ -43,7 +43,7 @@ class AvSelectField extends Component {
     let feedback = null;
     if (validation.errorMessage) {
       const feedbackClasses = classNames('d-block', feedbackClass);
-      feedback = <AvFeedback className={feedbackClasses}>{validation.errorMessage}</AvFeedback>;
+      feedback = <AvFeedback className={feedbackClasses} id={`${id}-error`}>{validation.errorMessage}</AvFeedback>;
```
jordan-a-young: Do you mind changing the suffix to be `-feedback` instead of `-error`. That way it will match the other library we have
https://github.com/Availity/availity-react/blob/master/packages/form/src/Feedback.js#L7

mdiallo21: @jordan-a-young , I wont. Changing it now

LauRoxx: I'm also not sure adding the id will help adding error information to this component since react-select's aria-describedby points to the placeholder and screen reader text.

Edit: oh wait, scratch that. I forgot you can use aria-errormessage.

mdiallo21: @LauRoxx , I'm not following you. I though the id added for the feedback can be referenced with in the AvSelectField by aria-describedby

mdiallo21: @LauRoxx  and @jordan-a-young , please review the new addition/changes

packages/reactstrap-validation-select/src/AvSelectField.js
```diff
@@ -41,9 +41,12 @@ class AvSelectField extends Component {
 
     const validation = this.context.FormCtrl ? this.context.FormCtrl.getInputState(name) : {};
     let feedback = null;
+    let ariaFeedback = {};
     if (validation.errorMessage) {
       const feedbackClasses = classNames('d-block', feedbackClass);
-      feedback = <AvFeedback className={feedbackClasses}>{validation.errorMessage}</AvFeedback>;
+      const feedbackId = `${id}-feedback`;
+      feedback = <AvFeedback className={feedbackClasses} id={feedbackId}>{validation.errorMessage}</AvFeedback>;
+      ariaFeedback = {hasError: true, ariaErrorId: feedbackId, ariaErrorDesc: validation.errorMessage}
```
LauRoxx, line 49: Why are we hardcoding hasError to true? feels like it defeats the purpose of having a variable. the logic for touched and hasError should already be available in avSelect that you can use instead of passing a boolean. 

Also, is there a reason why the names of the variables are constantly changing? for consistency I would rename ariaErrorId as feedbackId, ariaErrorDesc to errorMessage, and pick whether you want ariaFeedback or ariaMessage and keep it consistent between both AvSelect and AvSelectField

packages/reactstrap-validation-select/src/AvSelect.js
```diff
@@ -259,6 +261,16 @@ class AvSelect extends AvBaseInput {
       Tag = Creatable;
     }
 
+     // Enhance placeholder for accessibility
+  placeholder = (
+    <>
+      {placeholder || 'Select...'}
+      <span className="sr-only">
+        {ariaMessage.ariaErrorDesc || null} {helpMessage || null}
```
LauRoxx, line 269: - we don't want an error message to appear until the field has been touched (the feedback doesn't appear until after you leave the field for the first time, so it's the same), so check if touched and hasError . 
- Since ariaMessage isn't required I'd make it `ariaMessage?.ariaErrorDesc`

packages/reactstrap-validation-select/src/AvSelect.js
```diff
@@ -360,6 +372,9 @@ class AvSelect extends AvBaseInput {
             primary: '#3262af',
           },
         })}
+        aria-errormessage={ariaMessage.hasError ? ariaMessage.ariaErrorId : ''}
+        aria-invalid={ariaMessage.hasError}
```
LauRoxx, line 376: use the touched and hasError vars in this component to check for invalid state.

packages/reactstrap-validation-select/src/AvSelect.js
```diff
@@ -360,6 +372,9 @@ class AvSelect extends AvBaseInput {
             primary: '#3262af',
           },
         })}
+        aria-errormessage={ariaMessage.hasError ? ariaMessage.ariaErrorId : ''}
```
LauRoxx, line 375: - use the touched and hasError vars in this component to check for invalid state.
- We don't want the message to show in the errormessage and placeholder simultaneously. do not add feedbackid unless it's invalid AND a selection has been made (since that is when the placeholder is no longer there)

packages/reactstrap-validation-select/types/AvSelect.d.ts
```diff
@@ -5,6 +5,7 @@ export interface AvSelectProps extends AvInput {
   loadOptions?: Function;
   raw?: boolean;
   autofill?: boolean | object;
+  ariaMessage?: Object;
```
LauRoxx, line 8: we need to define what this object looks like

LauRoxx (changes requested):
Once you're done with this make sure to add the new props to the docusaurus documentation for these components.

mdiallo21:
Not needed

PR #1155, opened 2022-09-23
Updates and export some more utils.

danielpickett:
updating some types and changing some parameter names after feedback from those who know more than I do about SSOs. Also directly exporting updateTopApps.

### DIFF
--- a/packages/spaces/src/linkHandlers.js
+++ b/packages/spaces/src/linkHandlers.js
@@ -3,45 +3,38 @@ import nativeForm from '@availity/native-form';
 import { isAbsoluteUrl } from '@availity/resolve-url';
 import { updateTopApps, updateUrl } from './helpers';
 
-export const openLink = async (space, { akaname, linkAttributes }) => {
+export const openLink = async (space, { akaname, payerSpaceId }) => {
   if (!space?.link?.url) {
     return;
   }
 
-  await updateTopApps(space.configurationId, space.type, akaname);
+  await updateTopApps(space, akaname);
 
   const target = getTarget(space.link.target);
   const url = !isAbsoluteUrl(space.link.url)
     ? getUrl(
-        updateUrl(
-          space.link.url,
-          'spaceId',
-          space.parents && space.parents[0] ? space.parents[0].id : linkAttributes.spaceId
-        ),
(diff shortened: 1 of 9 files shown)


packages/spaces/src/linkHandlers.js
```diff
@@ -3,45 +3,38 @@ import nativeForm from '@availity/native-form';
...
 
   const target = getTarget(space.link.target);
   const url = !isAbsoluteUrl(space.link.url)
     ? getUrl(
-        updateUrl(
-          space.link.url,
-          'spaceId',
-          space.parents && space.parents[0] ? space.parents[0].id : linkAttributes.spaceId
-        ),
+        updateUrl(space.link.url, 'spaceId', space.parents && space.parents[0] ? space.parents[0].id : payerSpaceId),
```
jordan-a-young: I think we can change the last arg to be a little shorter with
```js
space.parents?.[0]?.id || payerSpaceId
```
what do you think?

danielpickett: You so fancy 😍 . I'll update it.

packages/spaces/src/linkHandlers.js
```diff
@@ -3,45 +3,38 @@ import nativeForm from '@availity/native-form';
...
       )
     : space.link.url;
+
   window.open(url, target);
 };
 
-export const openSsoLink = async (space, { akaname, clientId, linkAttributes, event }) => {
+export const openLinkWithSso = async (space, { akaname, clientId, payerSpaceId, ssoParams }) => {
   if (space.metadata && space.metadata.ssoId) {
-    event.preventDefault();
```
jordan-a-young, line 30: Will this be a breaking change if someone uses it on an anchor tag?

jordan-a-young: nevermind, just saw where you are using `event.preventDefault` later

danielpickett: Thanks for bringing this up. I went back to the diff of [my original change on Sept 19](https://github.com/Availity/availity-react/commit/f061e40039a1079eafad8ce08772c206d563f5e0) and I think I'm ok. Originally the onKeyPress would check for charCode 13, then check for metadata.ssoId (again) and then prevent default. Which is what the new code is doing in the same order if I'm reading it right.

jordan-a-young (approved):

gregmartDOTin (approved):

PR #1157, opened 2022-09-26
Add icon to feedback when in error

Valar05:

### DIFF
--- a/packages/form/src/Feedback.js
+++ b/packages/form/src/Feedback.js
@@ -3,19 +3,20 @@ import PropTypes from 'prop-types';
 import { FormFeedback } from 'reactstrap';
 import { ErrorMessage } from 'formik';
 
+const AvFeedback = ({ children, ...rest }) => (
+  <FormFeedback {...rest}>
+    <i className="icon icon-attention" aria-label="Error" />
+    {children}
+  </FormFeedback>
+);
+
 const Feedback = ({ name, ...rest }) => {
   const feedbackId = `${name}-feedback`.toLowerCase();
-  const AvFeedback = ({ children, ...rest }) => (
-    <FormFeedback {...rest}>
-      <i className="icon icon-attention" aria-label="Error" />
-      {children}
-    </FormFeedback>
-  );
-
-  AvFeedback.propTypes = {
-    children: PropTypes.node,
-  };
   return <ErrorMessage id={feedbackId} component={AvFeedback} name={name} {...rest} />;
+};
+
+AvFeedback.propTypes = {
+  children: PropTypes.node,
 };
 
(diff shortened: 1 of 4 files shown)


packages/form/src/Feedback.js
```diff
@@ -5,7 +5,17 @@ import { ErrorMessage } from 'formik';
 
 const Feedback = ({ name, ...rest }) => {
   const feedbackId = `${name}-feedback`.toLowerCase();
-  return <ErrorMessage id={feedbackId} component={FormFeedback} name={name} {...rest} />;
+  const AvFeedback = ({ children, ...rest }) => (
+    <FormFeedback {...rest}>
+      <i className="icon icon-attention" aria-label="Error" />
```
LauRoxx: looking back I don't believe aria-label is allowed on this element after all. it is probably better to make the icon aria-hidden and then add an sr-only span for the "Error" text.

packages/form/src/Feedback.js
```diff
@@ -5,7 +5,17 @@ import { ErrorMessage } from 'formik';
 
 const Feedback = ({ name, ...rest }) => {
   const feedbackId = `${name}-feedback`.toLowerCase();
-  return <ErrorMessage id={feedbackId} component={FormFeedback} name={name} {...rest} />;
+  const AvFeedback = ({ children, ...rest }) => (
+    <FormFeedback {...rest}>
```
nylon22: should this be prop driven? as in, will all web applications want to show an icon in the error message? 

and if so, what if a team wanted to show `icon-attention-circle` instead of `icon-attention` ? Do we have this locked down by design to drive consistency or should this component allow for more flexibility?

packages/form/src/Feedback.js
```diff
@@ -5,7 +5,17 @@ import { ErrorMessage } from 'formik';
...
+      <i className="icon icon-attention" aria-label="Error" />
+      {children}
+    </FormFeedback>
+  );
+
+  AvFeedback.propTypes = {
```
LauRoxx: We should move AvFeedback out of Feedback so we can do the proptypes outside of it as well, and it just looks nicer when all the proptypes are together.

packages/form/src/Feedback.js
LauRoxx: I believe we'd want it to be consistent across the portal. Teresa is the one that picked the icon. 

**also, I'd be worried the icon-attention-circle is a little too close to the help icon. since these icons are so tiny i think the difference in shape really adds to distinguishing them for colorblind people (which is the reason for the icon).

LauRoxx (approved):

nylon22 (approved):

PR #1161, opened 2022-10-10
@availity/table: Fix Accessibility issues and add DefaultValue support to formatted cells

smashleez:
This PR addresses the following:
- Accessibility issues that were discovered from AMP scans
- Add support to add DefaultValues to Formatted Cells


jordan-a-young (approved):
I like the switch to using `id` over `index`

PR #1163, opened 2022-10-12
feat(form-upload): add prop btnColor to Upload component

stevenviar:
We use the Upload component in one of our apps and we wanted to change the color to better match with everything else. There isn't currently a way to change this, so I thought we could add a prop to handle it while still leaving the current default.


jordan-a-young (approved):
Thanks for adding docs and types!

PR #1171, opened 2022-11-03
Upload & Form-Upload progressbar accessibility

LauRoxx:
- add accessible name to filerow uploadprogressbar
- add corresponding stories

### DIFF
--- a/packages/form-upload/src/FileRow.js
+++ b/packages/form-upload/src/FileRow.js
@@ -61,7 +61,7 @@ const FileRow = ({ onRemove, children, file }) => {
         </div>
       </td>
       <td className="align-middle" style={{ width: '45%' }}>
-        <UploadProgressBar upload={file} />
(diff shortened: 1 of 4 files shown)


packages/form-upload/src/FileRow.js
```diff
@@ -61,7 +61,7 @@ const FileRow = ({ onRemove, children, file }) => {
         </div>
       </td>
       <td className="align-middle" style={{ width: '45%' }}>
-        <UploadProgressBar upload={file} />
+        <UploadProgressBar upload={file} aria-label={`${file.file.name} upload`} />
```
jordan-a-young, line 64: should we add `progress` to the end of the label or is that implied somehow?

LauRoxx: the screenreader already reads the role of progressbar aloud and if traversing left/right between the cells also reads the column header of "upload progress"

jordan-a-young (approved):

PR #1176, opened 2022-11-14
feat(form-upload): set table role to presentation in FileList component for accessibility

stevenviar:

### DIFF
--- a/packages/form-upload/src/FileList.js
+++ b/packages/form-upload/src/FileList.js
@@ -20,7 +20,7 @@ const FileList = ({ files, children, onRemoveFile, ...rest }) => {
   }
 
   return files.length > 0 ? (
-    <Table size="sm" hover style={{ tableLayout: 'fixed' }} {...rest}>
+    <Table size="sm" role="presentation" hover style={{ tableLayout: 'fixed' }} {...rest}>
       <caption className="sr-only">List of files uploaded</caption>
       <thead>
         <tr>
--- a/packages/upload/src/FileList.js
+++ b/packages/upload/src/FileList.js
@@ -26,7 +26,7 @@ const FileList = ({ files, children, onRemoveFile, onPasswordSubmit, passwordMod
   }
 
   return files.length > 0 ? (
-    <Table size="sm" hover style={{ tableLayout: 'fixed' }} {...rest}>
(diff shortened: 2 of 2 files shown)


bjnewman:
I actually don't think this is an appropriate use of the role="presentation" which is supposed to be used on tables that are used for layout. Here we have a table with the list of files uploaded and various properties of those files. Maybe I am missing something but I dont think every table should have this role.

stevenviar:
I was wondering if it was appropriate too, it isn't really clear to me when this should be used. In my opinion you could argue this table is used for layout because it isn't really a traditional table to present data, one of the columns is an icon to remove the upload. There are headers and they do correspond with what's in the table so I see how you can argue that it isn't too. This is getting flagged by accessibility tools as a violation, if it's a false positive though that's fine with me, we can leave it as is.

bjnewman:
I will leave it up to @LauRoxx to tell me if i'm wrong or if the UX for users with accessibility assisted tools is better with this role. I'm struggling to think what other html we would use to semantically represent this list if it isn't a table, apart from maybe a ul and li with a bunch of spans?

LauRoxx:
I agree with Ben, this table is a legitimate table. The rows and columns have purpose as opposed to just using it for style.
Continuum flags this as something that requires manual review to determine if it's actually an issue and in this case the table is correct.

**Sorry Steven, I thought I had answered your teams message forever ago and only just looked back to see that I hadn't.

stevenviar:
Manually testing this using VoiceOver for Mac the table seems unusable right now. None of the actual content inside the table is read just all of these headers. Adding the presentation role does let you actually navigate through the list of files but it isn't great either. Maybe the best solution is to fix this so it works correctly like a table? Unless other screen readers aren't having this issue, I have pretty limited experience using them.

LauRoxx:
It works correctly as a table for me using voiceover + chrome. Megan raised an issue in the 508 intercomm channel last week about the icon column since the icons are aria hidden and there's nothing else in those cells. While we should fix that all the other cells/columns/rows are working for me.

stevenviar:
Ah looks like it may just be an issue for Firefox, also seems to work for me in Chrome. Other tables seem to work fine for me in Firefox, so not sure what's different here. Well I'll close this then and thanks to both of you.

PR #1200, opened 2022-12-09
@availity/tree - Select All Link Should only display on the first Root Node.

smashleez:
This is a simple bug fix for the tree component. In RCM, the top root level tree has a 'Select All' link that selects all tree items.

In the React implementation of this, there was a check missing in the case where there are multiple top level root elements and the Select All Link was appearing twice.

This PR fixes that.

### DIFF
--- a/packages/tree/CHANGELOG.md
+++ b/packages/tree/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 This file was generated using [@jscutlery/semver](https://github.com/jscutlery/semver).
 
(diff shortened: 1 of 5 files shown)


jordan-a-young (dismissed):

smashleez:
@jordan-a-young Sorry, forgot to check to make sure I rebased commits that had [skip ci] on them so al the build processes weren't running. Should be good now.

jordan-a-young (approved):

PR #1223, opened 2023-01-10
fix/breadcrumbs: Property Export 'Crumb' Type

smashleez:
We have a use case for building dynamic breadcrumbs and we noticed that the type 'Crumb' was not exported from this package. This PR addresses that, as well as exports the BreadCrumbProps for in the future if we need to extend them in anyway.


jordan-a-young (dismissed):

smashleez:
@jordan-a-young I am not really sure what is happening with this PR in regards to code coverage. When I look it says the breadcrumbs package is at 100%, so it makes me think it is something else. Can you take a look when you have a second?

jordan-a-young:
I'm not sure why it does this. I'm assuming there is a setting I need to update. I can override and merge

PR #1224, opened 2023-01-11
feat(select): ResourceSelect - Add support to leverage a regular GET rather than defaulting to POST GET. 

smashleez:
Traditionally for Remote Dropdowns we utilize a GET call (or `query` on the API Resource). We do not utilize graphQL so by default this will do a POST GET utilizing the ResourceSelect in @availity/select. 

With a POST GET, even when APIs are pagination on the server side, we are seeing it doubles the load time. After doing some research, this is common for POST GET to be slower due to sending the Request in the body rather than through query parameters. We thought we would be okay with the POST GET because it still returns the same results but the performance degradation has proven to be a show stopper. 

To get around this without impacting others, I updated the method prop to be able to conditionally take in 'GET' rather than 'POST' and when method = 'GET' call `query` on the API Resource.

### DIFF
--- a/docusaurus/docs/form/select/components/resource-select.md
+++ b/docusaurus/docs/form/select/components/resource-select.md
@@ -59,9 +59,9 @@ Configuration object used in the query method on the resource. Useful for defini
 
 Object used to create querystring parameters in the request. If function, will return new object with params for request.
 
-#### `method?: string`
+#### `method?: POST | GET`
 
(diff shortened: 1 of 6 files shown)


smashleez:
@jordan-a-young This Code Coverage failure is also peculiar on this PR. Did code coverage rules change? All of my changes are covered under unit tests, it seems to be complaining about other areas of the code that were not changed here.

jordan-a-young (approved):
Looks great!

PR #1229, opened 2023-01-17
Revert "fix(app-tiles): fixing version publish issue manually"

JoeSpanbauer:


jordan-a-young (changes requested):
We shouldn't manually touch the version. Instead you should add a change like a comment to bump the version to 0.1.3

PR #1242, opened 2023-02-14
Select package documentation

gk10:
Updated storybook to add documentation for the select package
descriptions taken from https://availity.github.io/availity-react/form/select/components/select


packages/select/stories/SelectField.stories.tsx
```diff
@@ -14,7 +15,10 @@ export default {
   title: 'Form Components/Select',
   parameters: {
     docs: {
-      // page: README,
+      description: {
+        component:
+          'The same as Select but with a Label that appears above the input and a Feedback that appears below the input.',
```
LauRoxx: The pattern we've been using is for the top level description to be the description for the entire package. Unfortunately with the way these stories are currently setup it seems this will list the selectfield description right underneath the package title ("select") making it a bit confusing.

I think it'd be best to move the selectfield story to the select.stories file and then add the description to the story itself ( see https://github.com/Availity/availity-react/blob/master/packages/form/src/Form.stories.tsx#L154-L159 ). Then we could get rid of this extra stories file.

Also, it'd be more accurate to say "Select with a Label and Feedback" as opposed to "the same as select but..." since it uses the select component.

packages/select/stories/SelectField.stories.tsx
```diff
@@ -112,3 +116,11 @@ export const SelectFieldStory: Story = ({
 );
 
 SelectFieldStory.storyName = '<SelectField />';
+
+export const SelectFieldProps: Story = () => (
```
LauRoxx: combine with the selectprops story once they're all in one file. 

If you run into any issues sometimes the use of multiple story names will interfere with the argstable. If that happens we should move away from the tag syntax for the story names and instead use some thing like selectField, _SelectField, etc and let storybook calculate the displayed story name.

packages/select/src/ResourceSelect.js
```diff
@@ -306,9 +306,9 @@ const ResourceSelect = ({
 };
 
 ResourceSelect.propTypes = {
+  /** The name of the field. Will be the key of the selected date that comes through in the values of the onSubmit callback. */
   name: PropTypes.string.isRequired,
-  requestConfig: PropTypes.object,
-  method: PropTypes.string,
+  /** Availity API resource (see @availity/api-axios). */
```
LauRoxx: I believe the argstable recognizes markdown notation if we want to insert a link for @availity/api-axios

packages/select/src/ResourceSelect.js
```diff
@@ -317,35 +317,62 @@ ResourceSelect.propTypes = {
...
+  method: PropTypes.string,
+  /** When a function, the function is called with the response body from the API call and is expected to return an array. When a string, the string is expected to be a simple key used to get the value from the response. ("simple" means dot notation is not supported for grabbing values from nested objects. If your result is deeply nested, provide a function.) */
   getResult: PropTypes.oneOfType([PropTypes.string, PropTypes.func]),
+  /** If true, ResourceSelect attempts to retrieve the next page of results. response.data from the axios response is passed as the only argument to hasMore when hasMore is a function. Defaults to: ({ totalCount, limit, offset }) => totalCount > offset + limit; for non-GraphQL apis. Defaults to (data) => data.data[${this.props.graphqlConfig.type}Pagination].pageInfo.hasNextPage for GraphQL apis. */
   hasMore: PropTypes.oneOfType([PropTypes.bool, PropTypes.func]),
+  /** Set to debounceTimeout if debounceTimeout is not provided. (see react-select-async-paginate) */
```
LauRoxx: insert a link (if possible)

packages/select/src/ResourceSelect.js
```diff
@@ -317,35 +317,62 @@ ResourceSelect.propTypes = {
...
   requiredParams: PropTypes.array,
+  /** Provide a list of properties to listen to from the parameters prop. If present, the options reset when any of the parameters specified in the array change value. This is useful for when a customerId changes and you need to load a new list of options for the user to choose from. This list will be used to derive cacheUniq when the cacheUniq prop is not provided. When using watchParams, the parameters prop must must be populated with values that are in the watchParams object. */
   watchParams: PropTypes.array,
+  /** The minimum number of characters the user must input before ResourceSelect makes the network request. If the user has not inputted any characters, the network request will still be made. Useful for relieving pressure on the api the resource is calling. */
   minCharsToSearch: PropTypes.number,
+  /** When this prop changes, all cached options are cleared. (see react-select-async-paginate) */
   // eslint-disable-next-line react/forbid-prop-types
   cacheUniq: PropTypes.any,
+  /** Additional properties to pass to AsyncPaginate (see react-select-async-paginate) */
```
LauRoxx: insert a link (if possible)

LauRoxx (approved):

PR #1247, opened 2023-02-17
docs(reactstrap-validation-date): reactstrap-validation-date storyboo…

gk10:
reactstrap-validation-date storybook props

### DIFF
--- a/packages/reactstrap-validation-select/src/AvResourceSelect.js
+++ b/packages/reactstrap-validation-select/src/AvResourceSelect.js
@@ -231,13 +231,13 @@ AvResourceSelect.create = (defaults) => {
 };
 
 AvResourceSelect.propTypes = {
-  requestConfig: PropTypes.object,
-  method: PropTypes.string,
   resource: PropTypes.shape({
     postGet: PropTypes.func,
     post: PropTypes.func,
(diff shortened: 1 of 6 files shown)


packages/reactstrap-validation-select/src/AvResourceSelect.js
```diff
@@ -231,13 +231,13 @@ AvResourceSelect.create = (defaults) => {
...
 AvResourceSelect.propTypes = {
-  requestConfig: PropTypes.object,
-  method: PropTypes.string,
   resource: PropTypes.shape({
     postGet: PropTypes.func,
     post: PropTypes.func,
     getResult: PropTypes.oneOfType([PropTypes.string, PropTypes.func]),
   }).isRequired,
+  requestConfig: PropTypes.object,
+  method: PropTypes.string,
```
LauRoxx, line 240: Is there a specific reason these are being moved? (I don't really touch the legacy components much so I'm not too familiar)

gk10: I was asked to have required props be at the top so I moved these below resource

LauRoxx: ahh, my window was tiny so it made it look like it was being moved into resource. whoops

LauRoxx (approved):

PR #1257, opened 2023-03-10
Feat/secrets

bjnewman:
Some additional secret detection added to our pre-push hook with an update to the contributing guide.


bjnewman:
Theres definitely an argument to be made that this should not be merged, as Chris points out we might be faced with support requests for installing a python dep which is probably outside of our skill set. But I'll leave it up for now as a draft for discussion until we make a decision.

PR #1266, opened 2023-04-14
Feat/select - Make Clear Button More Configurable

smashleez:
Ever since the Select components were updated to hard code the text 'clear' we have been asked my product and users to change this to be 'Clear' to make it more consistent with other buttons. 

With this change we can leave it default to be 'clear' to avoid impact on other areas where this might be expected, but allows us to customize the text. In addition we have a requirement to disable the button whenever there is no option selected, and that is facilitated by the addition of the `clearButtonProps`. 

Again, this just allows us to have more control over the Clear button other than just the 'class' of the button.

### DIFF
--- a/docusaurus/docs/form/select/components/select.md
+++ b/docusaurus/docs/form/select/components/select.md
@@ -128,9 +128,9 @@ Class names to add to clear button (only available when `isMulti` or `isClearabl
 
 Text that should be displayed in the clear button (only available when `isMulti` or `isClearable`). **Default:** `clear`
 
-#### `disableClearButtonWhenNoValue?: boolean`
+#### `clearButtonProps?: ButtonProps`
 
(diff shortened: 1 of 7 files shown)


LauRoxx (commented):
wondering if the disabling of the clear when there's no input should just be default behavior.

jordan-a-young (dismissed):

smashleez:
Yeah, I don't disagree - in some of our custom inputs we do default it to that behavior. Just didn't want to impact others whom might not want that behavior, but that is something they could always override as well.

smashleez:
@jordan-a-young @LauRoxx - I went ahead and updated the clear button to be disabled by default when there is no value. I feel like this should be expected behavior, but implementers should still be able to override this with the clearButtonProps if needed. 

Please rereview/approve.

LauRoxx (approved):

PR #1270, opened 2023-04-20
fix(spaces): useLink don't show multipayermodal if spaceId is passed

christieb7007:
When using SpacesLink, the MultiPayerModal is shown if the link config has multiple payerspace parents, which will send the selected payerspace's id in the URL. If the `spaceId` is already passed in the `linkAttributes`, use that spaceId in the URL rather than asking for it to be chosen.


jordan-a-young:
Looks like the unit tests need to be updated to account for the change. Otherwise the change looks good to me.

christieb7007:
@jordan-a-young I've made a couple updates, could you review?

jordan-a-young (approved):

PR #1278, opened 2023-05-09
Disabled prop #1277

### LINKED ISSUE #1277: Missing `disabled` prop type in PropTypes and form-upload.d.ts types file
### Current Behavior

I am noticing a lint error when I add a disabled prop to the FilePickerBtn from @availity/form-upload. There is a disabled prop in the [docs](https://availity.github.io/availity-react/form/upload/file-picker-btn#disabled-boolean), but not in the corresponding [proptypes](https://github.com/Availity/availity-react/blob/master/packages/form-upload/src/FilePickerBtn.js#L47-L62), or [FilePickerBtn.d.ts](https://github.com/Availity/availity-react/blob/master/packages/form-upload/src/FilePickerBtn.d.ts#L1C1-L11) file.

### Expected behavior

There should not be a lint error displayed. 

### Reproducible example

```jsx
<FilePickerBtn
    disabled <-- this prop throws lint error
    name="samplename"
    onChange={e => console.log(e)}
>
  Button
</FilePickerBtn>
```

### Suggested solution(s)

Add the disabled prop to https://github.com/Availity/availity-react/blob/master/packages/form-upload/src/FilePickerBtn.js#L47-L62 and https://github.com/Availity/availity-react/blob/master/packages/form-upload/src/FilePickerBtn.d.ts#L1C1-L11
### Your environment

  Binaries:
    Node: 16.15.0 - ~/.nvm/versions/node/v16.15.0/bin/node
    Yarn: 1.22.18 - ~/.homebrew/bin/yarn
    npm: 8.5.5 - ~/.nvm/versions/node/v16.15.0/bin/npm
  npmPackages:
    @availity/api-axios: ^7.0.3 => 7.3.1 
    @availity/api-core: ^8.0.4 => 8.0.4 
    @availity/authorize: ^2.2.0 => 2.2.1 
    @availity/badge: ^1.3.1 => 1.3.2 
    @availity/breadcrumbs: ^3.3.1 => 3.3.3 
    @availity/date: 3.4.4 => 3.4.4 
    @availity/favorites: ^3.4.2 => 3.4.3 
    @availity/form: ^1.7.3 => 1.7.3 
    @availity/form-upload: ^0.9.1 => 0.9.2 
    @availity/help: ^2.2.0 => 2.2.0 
    @availity/hooks: ^3.3.1 => 3.3.2 
    @availity/icon: ^0.11.1 => 0.11.2 
    @availity/link: ^3.0.0-beta.72 => 3.0.0-beta.72+a1f13030 
    @availity/list-group: ^1.4.1 => 1.4.4 
    @availity/list-group-item: ^1.4.1 => 1.4.3 
    @availity/localstorage-core: ^3.0.0 => 3.0.0 
    @availity/mock-server: ^7.2.1 => 7.2.1 
    @availity/page-header: ^11.3.1 => 11.3.1 
    @availity/pagination: ^2.15.2 => 2.15.5 
    @availity/phone: ^2.1.1 => 2.1.1 
    @availity/progress: ^1.5.1 => 1.5.1 
    @availity/select: ^3.3.2 => 3.4.0 
    @availity/spaces: ^6.3.0 => 6.5.7 
    @availity/training-link: 1.6.0 => 1.6.0 
    @availity/typography: ^1.4.1 => 1.5.0 
    @availity/upload-core: ^5.0.3 => 5.0.3 
    @availity/workflow: ^9.2.4 => 9.2.4 
    @availity/workflow-babel-preset: ^6.1.1 => 6.1.1 
    @availity/yup: ^5.1.2 => 5.1.2 

```bash
```

clintonlunn:

### DIFF
--- a/packages/form-upload/src/Upload.stories.tsx
+++ b/packages/form-upload/src/Upload.stories.tsx
@@ -2,9 +2,9 @@ import React from 'react';
 import { Meta, Story } from '@storybook/react';
 import { Form } from '@availity/form';
 
-import Upload, { FileList, FilePicker, FilePickerBtn, UploadProgressBar } from '.';
 import { ArgsTable } from '@storybook/addon-docs';
 import { CustomInput } from 'reactstrap';
+import Upload, { FileList, FilePicker, FilePickerBtn, UploadProgressBar } from '.';
 
 const file = (fileName: string) => ({
   id: fileName,
--- a/packages/form-upload/src/FilePicker.d.ts
+++ b/packages/form-upload/src/FilePicker.d.ts
@@ -1,13 +1,13 @@
 import React from 'react';
 
-export interface FilePickerProps<T extends React.ElementType> {
-  tag?: T;
+export interface FilePickerProps {
+  tag?: React.ElementType;
   multiple?: boolean;
   children?: React.ReactNode;
   name?: string;
   allowedFileTypes?: string[];
   maxSize?: number | string;
-  onChange?: React.ComponentProps<T>['onChange'];
+  onChange?: React.ChangeEventHandler<HTMLInputElement>;
 }
 
 declare const FilePicker: (props: FilePickerProps) => JSX.Element;
--- a/packages/form-upload/src/FilePicker.d.ts
+++ b/packages/form-upload/src/FilePicker.d.ts
@@ -1,13 +1,13 @@
 import React from 'react';
 
-export interface FilePickerProps {
-  tag?: React.ElementType;
-  onChange?: Function;
+export interface FilePickerProps<T extends React.ElementType> {
+  tag?: T;
   multiple?: boolean;
-  children?: Function;
+  children?: React.ReactNode;
   name?: string;
   allowedFileTypes?: string[];
   maxSize?: number | string;
+  onChange?: React.ComponentProps<T>['onChange'];
 }
 
 declare const FilePicker: (props: FilePickerProps) => JSX.Element;
--- a/packages/form-upload/src/FilePickerBtn.d.ts
+++ b/packages/form-upload/src/FilePickerBtn.d.ts
@@ -1,5 +1,4 @@
 export interface FilePickerBtnProps extends React.ButtonHTMLAttributes<HTMLButtonElement> {
-  tag?: React.ElementType;
   onChange?: React.ChangeEventHandler<HTMLInputElement>;
   multiple?: boolean;
   name?: string;
--- a/packages/form-upload/src/FilePickerBtn.d.ts
+++ b/packages/form-upload/src/FilePickerBtn.d.ts
@@ -1,10 +1,8 @@
-export interface FilePickerBtnProps {
-  onClick?: Function;
-  onChange?: Function;
+export interface FilePickerBtnProps extends React.ButtonHTMLAttributes<HTMLButtonElement> {
+  tag?: React.ElementType;
+  onChange?: React.ChangeEventHandler<HTMLInputElement>;
   multiple?: boolean;
   name?: string;
-  color?: string;
-  children?: React.ReactType;
   allowedFileTypes?: string[];
   maxSize?: number;
   'data-testid'?: string;
--- a/packages/form-upload/tests/FilePickerBtn.test.js
+++ b/packages/form-upload/tests/FilePickerBtn.test.js
@@ -24,4 +24,22 @@ describe('FilePickerBtn', () => {
     );
     expect(container).toBeDefined();
   });
+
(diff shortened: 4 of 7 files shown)


clintonlunn:
I added the same props to both `form-upload` and `upload` packages. It looks like what was broken in my example too. I haven't updated the stories just yet, but I can with whatever needs to be put in there.

bjnewman:
isn't this the same issue for any html property? i think the interfaces should extend the base React button interface to inherit all those properties? See https://mortenbarklund.com/blog/react-typescript-props-spread/ for example

clintonlunn:
> isn't this the same issue for any html property? i think the interfaces should extend the base React button interface to inherit all those properties? See https://mortenbarklund.com/blog/react-typescript-props-spread/ for example

Thanks for the comment. That makes sense. In that case, I also get errors about onClick and onChange,I am assuming we would also want to pass those along from a base type.

image


i.e. maybe we'd want to do something like this:
```jsx
import type { ButtonHTMLAttributes, ComponentPropsWithoutRef } from 'react';

export interface FilePickerBtnProps extends ComponentPropsWithoutRef<'button'> {
  multiple?: boolean;
  name?: string;
  color?: string;
  children?: React.ReactType;
  allowedFileTypes?: string[];
  maxSize?: number;
  'data-testid'?: string;
}
```

or 
```jsx
import type { ButtonHTMLAttributes, ComponentPropsWithoutRef } from 'react';

export interface FilePickerBtnProps extends ButtonHTMLAttributes<HTMLButtonElement> {
  multiple?: boolean;
  name?: string;
  color?: string;
  children?: React.ReactType;
  allowedFileTypes?: string[];
  maxSize?: number;
  'data-testid'?: string;
}
```

jordan-a-young (dismissed):
🚀

clintonlunn:
I can't quite get the type for onChange to pass through for the FilePicker component unless i do something like `export interface FilePickerProps<T extends React.ElementType = 'input'> {...` where i define a default react elementtype or if i make the on change something like  ` onChange?: React.ChangeEventHandler<HTMLInputElement>`. It seems like the tag can be a number of different types of elements, so passing a correct onchange would mean that we would want to keep as generic as possible?

bjnewman:
> I can't quite get the type for onChange to pass through for the FilePicker component unless i do something like `export interface FilePickerProps<T extends React.ElementType = 'input'> {...` where i define a default react elementtype or if i make the on change something like ` onChange?: React.ChangeEventHandler<HTMLInputElement>`. It seems like the tag can be a number of different types of elements, so passing a correct onchange would mean that we would want to keep as generic as possible?

the onChange gets passed to the FilePicker so it should have the same type as FIlePicker's onChange which is, yeah a little wider than React.ChangeEventHandler\<HTMLInputElement> since the input tag can be overridden. Does React.ChangeEventHandler<HTMLElement | React.ComponentType> or something similarly broad work?

bjnewman:
> > I can't quite get the type for onChange to pass through for the FilePicker component unless i do something like `export interface FilePickerProps<T extends React.ElementType = 'input'> {...` where i define a default react elementtype or if i make the on change something like ` onChange?: React.ChangeEventHandler<HTMLInputElement>`. It seems like the tag can be a number of different types of elements, so passing a correct onchange would mean that we would want to keep as generic as possible?
> 
> the onChange gets passed to the FilePicker so it should have the same type as FIlePicker's onChange which is, yeah a little wider than React.ChangeEventHandler<HTMLInputElement> since the input tag can be overridden. Does React.ChangeEventHandler<HTMLElement | React.ComponentType> or something similarly broad work?

After talking to Clinton I think something more like 
image
should be good enough for our onChange type without getting too generic

packages/form-upload/src/Upload.stories.tsx
```diff
@@ -2,9 +2,9 @@ import React from 'react';
 import { Meta, Story } from '@storybook/react';
 import { Form } from '@availity/form';
 
-import Upload, { FileList, FilePicker, FilePickerBtn, UploadProgressBar } from '.';
 import { ArgsTable } from '@storybook/addon-docs';
 import { CustomInput } from 'reactstrap';
+import Upload, { FileList, FilePicker, FilePickerBtn, UploadProgressBar } from '.';
```
clintonlunn, line 7: lint error told me to move it down here

bjnewman (approved):
🔥 thank you

clintonlunn:
@bjnewman i coudn't get your suggestion above to type things correctly so i went with a more simplistic approach of declaring only the onchange as a `React.ChangeEventHandler<HTMLInputElement>` and everything else will get either custom typed or inherited

PR #1284, opened 2023-05-30
fix(date): add placeholder and additional format option

henten20:

### DIFF
--- a/packages/date/src/DateField.d.ts
+++ b/packages/date/src/DateField.d.ts
@@ -10,6 +10,7 @@ export type DateFieldProps = {
   labelAttrs?: LabelProps;
   required?: boolean;
   helpId?: string;
+  placeholder?: string;
 } & DateProps;
 
 declare const DateField: (props: DateFieldProps) => JSX.Element;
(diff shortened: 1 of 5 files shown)


packages/date/src/DateField.js
```diff
@@ -4,7 +4,18 @@ import { FormGroup, Feedback, Label } from '@availity/form';
...
+const DateField = ({
+  name,
+  label,
+  labelClass,
+  labelHidden,
+  labelAttrs,
+  id = name,
+  required,
+  helpId,
+  placeholder,
```
jordan-a-young: You don't need to pull out `placeholder` here. It will stay as part of the `props` that get spread on line 32

packages/date/src/DateField.d.ts
```diff
@@ -10,6 +10,7 @@ export type DateFieldProps = {
   labelAttrs?: LabelProps;
   required?: boolean;
   helpId?: string;
+  placeholder?: string;
```
jordan-a-young: Sorry, missed this before. This is extending the DateProps which already has the `placeholder` prop so we don't need to add it here

jordan-a-young (approved):

PR #1299, opened 2023-07-28
feat(spaces): add showName prop, special handling for type FILE to Sp…

nylon22:
…acesLink


chrishavekost (dismissed):

packages/spaces/src/SpacesLink.js
```diff
@@ -259,7 +261,13 @@ const Link = ({
       >
         {!stacked && favoriteIcon}
         {appIcon}
-        {icon && <Icon name={icons.navigation} />}
+        {icon && type.toUpperCase() === 'FILE' ? (
+          <AvLink target="_blank" href={restLink.url}>
```
chrishavekost, line 265: I don't think the Codecov diff is preventing a merge, but if you have time can you add a test for this new logic?

jordan-a-young (dismissed):

PR #1307, opened 2023-09-01
fix(select): updating react-select and adding aria-controls to input …

JoeSpanbauer:
…to fix 508 compliance issues EB-234

### DIFF
--- a/packages/select/src/Select.js
+++ b/packages/select/src/Select.js
@@ -25,7 +25,7 @@ const components = {
   ClearIndicator: () => null,
   Input: (props) => {
     const { 'aria-required': required } = props.selectProps;
-    return <Input {...props} aria-required={required} />;
+    return <Input aria-controls="" {...props} aria-required={required} />;
   },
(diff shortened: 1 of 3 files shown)


packages/select/src/Select.js
```diff
@@ -25,7 +25,7 @@ const components = {
   ClearIndicator: () => null,
   Input: (props) => {
     const { 'aria-required': required } = props.selectProps;
-    return <Input {...props} aria-required={required} />;
+    return <Input aria-controls="" {...props} aria-required={required} />;
```
gregmartDOTin, line 28: I think we need to be able to pass a value for aria-controls, right?

JoeSpanbauer: The component we're using doesn't support this. aria-controls needs to be the id of the ul, and this package uses divs. Also, the menu isn't rendered until it's expanded, which doesn't resolve the 508 compliance issue.

owolfhu1 (approved):

PR #1316, opened 2023-10-13
docs(authorize): update stories for Authorize

ncruz-av:

### DIFF
--- a/packages/authorize/src/Authorize.stories.tsx
+++ b/packages/authorize/src/Authorize.stories.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { Story, Meta } from '@storybook/react';
+import { StoryObj } from '@storybook/react';
 import { Alert } from 'reactstrap';
 import BlockUi from '@availity/block-ui';
 import { QueryClientProvider, QueryClient } from 'react-query';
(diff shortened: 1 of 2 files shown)


ncruz-av:
The _UseAuthorize StoryObj has a typeof BlockUI, but the name of the story should be 'useAuthorize' right? That is why I left the storyName in it, but I wasn't sure if that was the correct thing.

packages/authorize/src/Authorize.stories.tsx
```diff
@@ -36,47 +36,51 @@ export default {
...
     </div>
-  );
+  ),
+  args: {
+    negate: false,
+    loader: true,
+  },
+};
+
+export const _UseAuthorize: StoryObj<typeof BlockUi> = {
```
LauRoxx: > The _UseAuthorize StoryObj has a typeof BlockUI, but the name of the story should be 'useAuthorize' right? That is why I left the storyName in it, but I wasn't sure if that was the correct thing.

afaik the type isn't super important if you're not relying on storybook to infer the args/controls, but since you have them explicitly listed I would assume it's alright. Does anything odd jump out when you view the story through the canvas (go to the canvas tab for the story to see the associated controls with the story)

LauRoxx: to answer your original question the `useAuthorize` name is correct since that's the hook/behavior you're trying to showcase.

ncruz-av: Nothing odd jumps out to me.

LauRoxx: same comment as feedback, each exported component/hook should be its own stories file

LauRoxx (changes requested):

LauRoxx (approved):

PR #1326, opened 2023-10-17
docs(feedback): update Storybook change for feedback

ncruz-av:


packages/feedback/stories/feedback.stories.tsx
```diff
@@ -1,156 +1,157 @@
...
 
 export default {
   title: 'Components/Feedback',
-  parameters: {
-    docs: {
-      // page: README,
-    },
-  },
-} as Meta;
+  components: { Feedback, FeedbackForm },
```
LauRoxx: every component should be it's own stories file so it gets its own generated props table. Just add to the title for each, so `Components/Feedback/Feedback` & `Components/Feedback/FeedbackForm`

LauRoxx (changes requested):

packages/feedback/stories/FeedbackWithModal.stories.tsx
```diff
@@ -0,0 +1,71 @@
+import React from 'react';
+import { StoryObj } from '@storybook/react';
+
+import Feedback from '../src/Feedback';
+import FeedbackForm from '../src/FeedbackForm';
+// import README from '../README.md';
+
+export default {
+  title: 'Components/Feedback/FeedbackWithModal',
+  components: Feedback,
```
LauRoxx: I think there was some confusion here. Feedback with modal still is based on the feedback component, so should stay with the other feedback stories. FeedbackForm is different since it's a separate componnet.

LauRoxx (approved):

PR #1345, opened 2023-10-27
docs(form): update stories to csf3 for Training Form  component

ncruz-av:


packages/form/src/Form.stories.tsx
```diff
@@ -35,7 +35,7 @@ import { Props as FormGroupProps } from './FormGroup';
 
 export default {
   title: 'Form Components/Form',
-  component: Form,
+  component: FormResults,
```
jordan-a-young: I think this should stay as `Form` since that is the component we publish.

jordan-a-young:
For each of the `StoryObj` types the type we pass in should match what the story is for since that type is used to generate docs

packages/form/src/CheckboxGroup.stories.tsx
```diff
@@ -0,0 +1,79 @@
...
+};
+
+type FormStoryProps = { required: boolean };
+
+type CheckboxStoryProps = {
+  helpId: string;
+  labelClassName: string;
+} & FormStoryProps;
+
+export const _Default: StoryObj<typeof CheckboxGroup> = {
```
LauRoxx: Have the story names match the component name, so for this it would be `_CheckboxGroup`

packages/form/src/Form.stories.tsx
```diff
@@ -383,6 +146,5 @@ export const hidden_Label = (props: LabelProps) => <Label {...props} />;
 export const hidden_Field = (props: FieldProps) => <Field {...props} />;
 export const hidden_Checkbox = (props: CheckboxProps) => <Checkbox {...props} />;
 export const hidden_CheckboxGroup = (props: CheckboxGroupProps) => <CheckboxGroup {...props} />;
-export const hidden_Radio = (props: RadioStoryProps) => <Radio {...props} />;
```
LauRoxx, line 386: additional props page breaks since this was removed. can we add it back along with adding radiogroup?

packages/form/src/CheckboxGroup.stories.tsx
```diff
@@ -0,0 +1,79 @@
...
+    },
+  },
+  args: {
+    required: false,
+  },
+};
+
+type FormStoryProps = { required: boolean };
+
+type CheckboxStoryProps = {
```
LauRoxx: are any of these story props needed?

packages/form/src/CheckboxGroup.stories.tsx
```diff
@@ -0,0 +1,79 @@
...
+  render: ({ required, helpId, labelClassName }) => (
+    <FormResults
+      onSubmit={() => {
+        console.log('submitted');
+      }}
+      initialValues={{
+        checkboxGroup: [],
+      }}
+      validationSchema={yup.object().shape({
+        checkboxGroup: yup.array().isRequired(required, 'At least one checkbox is required'),
```
LauRoxx: it looks like there's an alternative way to code the `isRequired` bit.

```js
validationSchema={yup.object().shape({
        checkboxGroup: yup.array().when([], {
          is: () => required,
          then: yup.array().required("At least one checkbox is required"),
          otherwise: yup.array().notRequired(),
        }),
      })}
````

so in general the schema would be

```js
validationSchema={yup.object().shape({
        {field name}: yup.{string/array/object}().when([], {
          is: () => required,
          then: yup.{string/array/object}().required("This field is required."),
          otherwise: yup.{string/array/object}().notRequired(),
        }),
      })}
```

packages/form/src/CheckboxGroup.stories.tsx
ncruz-av: They were not needed

LauRoxx (approved):

PR #1355, opened 2024-01-10
Feat/provider filter

bjnewman:
Filtering all providers unless devs pass in searchAll prop.


bjnewman:
i forgot to account for the fact that parameters can be a function, will update the PR in a bit to address this

bjnewman:
converting to draft - it turns out that more internal discussion is needed before this change can be made ;)

PR #1404, opened 2024-05-02
feat(form): added logic for changing default help icon to video icon

nicolas-0811:
With this change, we now have the ability to choose between the help icon or the video icon when using the FieldHelpIcon component

### DIFF
--- a/packages/form/src/Checkbox.d.ts
+++ b/packages/form/src/Checkbox.d.ts
@@ -10,6 +10,7 @@ export interface CheckboxProps extends HTMLAttributes<HTMLInputElement> {
   groupClassName?: string;
   groupName?: string;
   helpId?: string;
+  isHelpVideoType?: boolean;
 }
 
 declare const Checkbox: (props: CheckboxProps) => JSX.Element;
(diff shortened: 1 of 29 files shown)


jordan-a-young (changes requested):
I think this is a good start. I have a few requests to add

1. Can you update the [FieldHelpIcon.d.ts](https://github.com/Availity/availity-react/blob/master/packages/form/src/FieldHelpIcon.d.ts#L1) file to include the `isVideoType` prop?
2. Can you add the prop to [Field](https://github.com/Availity/availity-react/blob/master/packages/form/src/Field.js#L105) and [Label](https://github.com/Availity/availity-react/blob/master/packages/form/src/Label.js#L51) so that they get passed down to `FieldHelpIcon`? They will also need the prop added to their corresponding `.d.ts` files

packages/form/src/Checkbox.d.ts
```diff
@@ -10,6 +10,7 @@ export interface CheckboxProps extends HTMLAttributes<HTMLInputElement> {
   groupClassName?: string;
   groupName?: string;
   helpId?: string;
+  isVideoType?: boolean | false;
```
jordan-a-young: This can just be `isVideoType?: boolean`. `boolean` includes `false`

jordan-a-young: This happens in some other files and should be changed there as well

jordan-a-young (approved):

PR #1458, opened 2024-08-23
feat(link): update link to encode essentials urls

bjnewman:


packages/link/src/Link.js
```diff
@@ -4,8 +4,10 @@ import classNames from 'classnames';
...
 
 // if absolute or loadApp is disabled, return url. otherwise loadappify the url
-export const getUrl = (url = '', loadApp, absolute) => {
-  if (absolute || !loadApp) return url;
+export const getUrl = (url = '', loadApp = false, absolute = false) => {
+  const isEssentials = /(test|qa(p?)-)?essentials\.availity\.com/.test(url);
+
+  if ((absolute || !loadApp) && !isEssentials) return url;
 
   return `/public/apps/home/#!/loadApp?appUrl=${encodeURIComponent(url)}`;
```
LauRoxx, line 12: Oof, monday brain. Does having the relative link before loadApp still work if it's coming from an essentials page?

bjnewman: I made the PR with the intention to cover the the use case currently needed for urgent deliverables where apps accessed through apps.availity.com need to link to apps hosted on essentials.availity.com. I do think that this needs to be handled in a more systematic way, for the combination of cases for apps <-> essentials and essentials <-> essentials where now we only cover apps <-> apps.

LauRoxx (approved):

jordan-a-young (approved):

PR #1522, opened 2025-02-13
Chore/update upload core

jordan-a-young:
This upgrades @availity/upload-core for both `upload` and `form-upload`. The breaking changes should be minimal, but I still wanted to mark it as one in case.


packages/form-upload/src/FilePicker.js
```diff
@@ -33,7 +33,7 @@ const FilePicker = ({
     for (let i = 0; i < files.length; i++) {
       value[i] = files[i];
     }
-    setFieldValue(name, value, true);
+    // setFieldValue(name, value, true);
```
gregmartDOTin: Is this supposed to be commented out?

jordan-a-young: Yup, supposed to be deleted. Will fix

jordan-a-young: That was setting the context to a `File` instead of an `Upload` so I handle the update higher up

gregmartDOTin (approved):

LauRoxx (approved):

PR #1569, opened 2025-05-15
feat: drop support for node 16 add support for node 20 and 22

LauRoxx:


package.json
```diff
@@ -73,9 +73,9 @@
     "@babel/preset-react": "^7.26.3",
     "@babel/preset-typescript": "^7.27.0",
     "@jscutlery/semver": "^5.3.1",
-    "@nx/eslint": "19.5.7",
-    "@nx/jest": "19.5.7",
-    "@nx/workspace": "19.5.7",
+    "@nx/eslint": "^20.0.0",
+    "@nx/jest": "^20.0.0",
+    "@nx/workspace": "^20.0.0",
```
gregmartDOTin: I thought there were some migrations involved with going from nx v19->v20

gregmartDOTin (changes requested):

package.json
LauRoxx: I only saw generator stuff, but we don't use a generator in this repo. I'll take another look tho.

gregmartDOTin: Did you upgrade via their upgrade script? It'll do it all for you.

LauRoxx: I forgot that existed. I ran it and it only upgraded typescript. lol

gregmartDOTin (approved):

